PR #13, opened 2026-09-11
fix(pseudonymizer): treat agent tool blocks as text, not as attachments

### LINKED ISSUE #12: pseudonymizer: agent tool_result blocks are treated as unprocessable attachments, blocking Claude Code behind any pseudonymizing pipeline
## Summary

An agentic client (Claude Code, via the Anthropic Messages API) cannot use a virtual model whose pipeline contains a `pseudonymizer` node. Its `tool_result` blocks carry file content **by reference**, and the plugin treats every such part as an attachment it cannot pseudonymize. Both available policies then fail the run, for different reasons.

Tested on **0.65.0**, `pseudonymizer` 0.1.0, default node config.

## What happens

`attachmentText` rejects a part whose `Data` is empty (`plugin.go`, `reasonNoInlineData` — *"fichier transmis par référence, contenu illisible pour le filtre"*), so the part lands in `removedParts`. From there:

- **`unsupported_attachments: block`** (default) — the request is refused before the first turn:

  ```
  403 request rejected by pipeline: Requête refusée par le pseudonymiseur :
  2 pièces jointes ne peuvent pas être pseudonymisées
  - Skill : fichier transmis par référence, contenu illisible pour le filtre
  ```

- **`unsupported_attachments: remove`** — the request goes through, but every `tool_result` is stripped. The agent's read tools return nothing, it keeps going blind, and the run ends with no useful output. A warning is appended each turn, so it is not silent — but it is not actionable by the agent either.

- **`process_attachments: false`** does not help: `attachmentText` returns `reasonDisabled`, so the part is still removed.

## Evidence

Same merge request, same CI script, same runner; only the model changed:

| model | turns | outcome |
|---|---|---|
| virtual model with pseudonymizer, `block` | 2 | 403, nothing produced |
| same, `remove` | 20 | reads empty, nothing produced |
| the bare provider model it wraps | 108 | full review produced, 4 findings + summary |

## Why it matters

A pipeline that pseudonymizes is precisely what one wants in front of an agent that reads a private codebase — it is the case with the most to protect. Today that combination is the one that cannot work, so the practical answer is to route agents to a bare model, which is the opposite of the intent.

## What would help

Distinguishing a **`tool_result` produced by the client's own tools** from a **document attached by a human** would be enough. The former is text the model asked for and that the gateway can pseudonymize like any other message content; the latter is the case the attachment policy was designed for.

Failing that, a third policy value (`passthrough`, applying to tool traffic only) would let an operator keep `block` for human attachments while letting agent traffic through the text pseudonymizer.

Happy to test a patch against our instance.

scriptted:
Fixes #12.

A `tool_use` or `tool_result` block carries no inline file bytes, so it fell through to the attachment path, which declared it unreadable and applied the attachment policy. Both outcomes break an agentic client:

- `block` (the default) refuses the request before the first turn;
- `remove` strips the very blocks the agent asked for, and it keeps working blind — measured at 20 turns and no usable output.

Neither is what that policy was written for. It exists for a document a human attached and believes was analyzed. A `tool_result` is text the model itself requested, and the gateway can pseudonymize it like any other message content.

## What the patch does

Tool blocks are rewritten in place instead of being routed to the attachment path:

- `tool_result` — the `content`, whether a string or a list of text blocks;
- `tool_use` — the values inside `input`, at any depth.

What correlates a call with its result (`id`, `tool_use_id`, `name`) is left alone, and so are the argument *names*: renaming them would produce a call the client cannot execute.

A tool block carrying a **non-textual** payload — an image returned by a screenshot tool — has that payload replaced in place by a short notice addressed to the model, and keeps its position. It is never dropped: a `tool_result` is half of a pair, and removing it leaves the matching `tool_use` unpaired, which the Messages API answers with a 400. Mixed content keeps the text it could read.

Tool blocks therefore never reach `removedParts`, and the attachment policy does not apply to them at all — under `block` as under `remove`. That policy exists for a document a human attached and believes was analyzed, not for a pair the protocol needs intact.

## Measurements

On our instance, before the fix. Same merge request, same CI script, same runner; only the model changed:

| model | turns | outcome |
|---|---|---|
| virtual model with this node, `block` | 2 | 403, nothing produced |
| same, `remove` | 20 | reads empty, nothing produced |
| the bare provider model it wraps | 108 | full review, 4 findings + summary |

The pipeline that pseudonymizes was the only one that could not be put in front of an agent reading a private codebase — the case with the most to protect.

## Tests

`plugins/pseudonymizer/prerequest_tool_part_test.go`, five cases through the existing `preRequestWithParts` harness (regex anonymizer, no NER model, no network):

- a string-shaped `tool_result` survives and is pseudonymized;
- a list-shaped one has each text block pseudonymized, innocuous blocks kept;
- `tool_use` arguments are pseudonymized while `id`, `name` and argument names are preserved;
- a `tool_result` carrying an image still follows the attachment policy, and the refusal names why;
- `anonymizeLeaves` preserves shape, keys and non-string values.

`go test ./plugins/pseudonymizer/` and `go test ./internal/...` pass.

## What this does not fix

Two limits, both raised in review, both left to follow-ups:

- **Placeholders are never restored inside response tool calls** (#14). `pipeline_wrapped_client.go:41` passes only `msg.Content()` to the backward pass. The model reads pseudonymized arguments in its own history and derives new calls from them, so a placeholder can reach the client verbatim — `Read(path="/home/[PERSON_1]/notes.md")` against a path that does not exist. The limitation predates this patch, but the patch makes it likely rather than theoretical, since paths and patterns are exactly what it rewrites. The 108-turn row above does not cover it either: the bare model pseudonymizes nothing, so no placeholder ever had to come back.
- **Only the Messages format is covered** (#15). OpenAI-shaped tool arguments live in `tool_calls[].function.arguments`, outside `content`, and are not walked. The node documentation now says which format is protected rather than leaving an operator to assume.

`thinking` and `redacted_thinking` blocks hit the same wall as the one fixed here and are tracked separately (#16), along with the suggestion to narrow `default:` to the known attachment types rather than adding part types in front of it.


Bornholm:
Thanks for the issue, the patch, and especially for the numbers. The diagnosis is right, and I think the shape of the fix is right too. Build and tests pass here (`go test ./plugins/pseudonymizer/`, 105 tests).

On your open question at the end: I'd rather not add a `passthrough` policy value. That would ask the operator to configure their way out of what is really a misclassification in the code. A `tool_result` was never an attachment, and the config shouldn't have to say so. Your distinction by part type is the one I want.

Three things I'd like to talk about before merging, then some small stuff.

### Dropping a tool_result breaks the call/result pairing

The non-textual case still feeds `removedParts`, so a tool that returns an image reproduces issue #12 one step further along. Under `block` the whole agentic session is refused. Under `remove` the `tool_result` disappears while the matching `tool_use` stays in the previous assistant message, and the Messages API answers 400 on the unpaired id.

Replacing the block with a text block saying the content was removed keeps the pairing and tells the agent what happened, which is closer to what the rest of your patch does:

```json
{"type": "text", "text": "[non-textual content removed by the pseudonymizer]"}
```

Same for mixed content: `[text, image]` currently drops the text too, and the text was the part we could handle.

This one I'd like in this PR.

### Placeholders in tool calls are never restored

`pipeline_wrapped_client.go:41` only hands `msg.Content()` to the backward pass, so `deanonymize` sees the assistant's text and nothing else. After this patch the model reads pseudonymized arguments in its own history. When it derives a new call from what it read, say `Read(path="/home/[PERSON_1]/notes.md")` or a grep over a name, the placeholder goes out to the client verbatim and the client runs a call against a path that doesn't exist.

The limitation predates your patch, but the patch makes it likely rather than theoretical, because paths and patterns are exactly what it now rewrites. Your 108-turn row doesn't cover it either, since the bare model pseudonymizes nothing.

Restoring placeholders inside response tool calls is the real fix and it's outside this PR. I'd be happy with a paragraph in the PR body naming the limit and a follow-up issue.

### thinking blocks hit the same wall

There's no handling of `thinking` or `redacted_thinking` anywhere in the plugin, so they fall through to `default:`, `extractAttachment` finds no inline bytes, and we're back to `reasonNoInlineData`. Anyone running Claude Code with extended thinking will hit the 403 again right after this merges.

Which is the argument for going after `default:` rather than adding part types in front of it: an unknown part is currently presumed to be an unreadable attachment. If `default:` only routed the known attachment types (`file`, `input_file`, `document`, `image`) to `attachmentText` and left everything else alone, `isToolPart` would nearly disappear. One caveat if you look at it: a `thinking` block's `signature` must survive byte for byte, so `anonymizeLeaves` can't be pointed at it as is.

Separate PR, no objection to merging this one first.

### Smaller things

- `anonymizeToolPart` returns `(nil, true, err)` on failure, where `handled` means nothing and the caller ignores it. Two functions, or a sentinel `errNonTextToolPart`, would read better than a triple where one value is sometimes a lie.
- `partName` on a `tool_result` finds neither `name` nor `title`, so the rejection message names an empty block. `tool_use_id` would do as a fallback.
- `NoResponseRewrite` will be false on essentially every agentic turn now, so those sessions lose streaming. Expected, but worth a line in the node docs.
- `buildInstructionText` lists every placeholder in the system message. Tool traffic carries whole files, so the mapping can reach several hundred entities and that instruction is re-sent every turn. A cap would be prudent.
- When anonymization fails, the original block goes through in clear. That matches the existing text path and the `passthrough` event reports it, so it's consistent. The stake is just no longer one sentence, it's a file.

Nice work on the test file, by the way. Going through the existing `preRequestWithParts` harness with no NER model and no network is what made it quick to review.

Bornholm (changes requested):
See https://github.com/xolo-gateway/xolo/pull/13#issuecomment-5639100875

Bornholm:
Two more things after a second pass. Neither changes what I said above, but the first one does change how I read the title of this PR.

### OpenAI-shaped tool calls aren't covered

`grep -rn tool_calls plugins/pseudonymizer/` returns nothing, and the loop in `PreRequest` only looks at `msg["content"]`. For an OpenAI-compatible client the assistant's tool arguments live in `tool_calls[].function.arguments` as a JSON string, outside `content` entirely, so they go to the provider in clear while their Anthropic equivalent (`tool_use.input`) is now rewritten.

That predates your patch, but it means "agent tool blocks" holds for the Messages format only, and an operator reading the title will assume otherwise. Either handle `tool_calls` here, or say in the PR body and the plugin README which format is protected. I'd take the second, and a follow-up for the first.

### The new branches aren't tested

`attachmentConfig` never sets `UnsupportedAttachments`, so all five tests run under `block`. What that leaves unexercised is exactly what this patch adds: the `remove` path with a non-textual `tool_result` (the pairing problem from my first comment), the error path of `anonymizeToolPart`, a mixed `[]any{text, image}` list, and a `content` that is an object rather than a string or a list.

A `remove` case with an image `tool_result` would have caught the 400 before I did, so I'd start there.

### Smaller, related

In the `[]any` branch, a sub-element that isn't a `map[string]any` is copied through untouched and the block still returns `handled=true` (`tool_part.go:66-69`). A bare string in a `content` list isn't valid per the spec, so this is unlikely in practice, but it sits right next to the line where a non-`text` block returns `handled=false` and gets the policy. Same list, two opposite answers. Running a `string` element through `anonymize` and giving anything else `handled=false` would settle it.

scriptted:
Thanks — the 400 on the unpaired id is a good catch, and it makes the first version of this patch a half-fix: I removed a block the protocol needs in pairs and would have reproduced #12 one step further along.

Pushed in f44fb22.

### Dropping a tool_result breaks the pairing

Fixed the way you suggested. What cannot be read is replaced in place by `[non-textual content removed by the pseudonymizer]`, and the block keeps its position. Mixed content keeps the text: dropping it with the image threw away the half the plugin can actually protect.

The consequence is broader than the notice itself — tool blocks now never reach `removedParts`, so the attachment policy does not apply to them **under `block` either**, not just under `remove`. That felt right rather than conservative: the policy exists for a document a human attached and believes was analyzed, and a tool block is neither.

Two of your smaller points dissolved with it, so I did not address them separately:

- the `(nil, true, err)` triple is gone — nothing is left "unhandled", so the middle value had nothing left to lie about, and `anonymizeToolPart` now returns `(map[string]any, error)`;
- `partName` needs no `tool_use_id` fallback, since a tool block can no longer appear in a rejection message at all.

### The new branches aren't tested

You were right that all five ran under `block`. Added: a non-textual `tool_result` under `remove` (the 400) and under `block`, mixed `[text, image]`, a `content` that is an object, a bare string element, and the anonymizer failure path for all three shapes. The failure path goes through `anonymizeToolPart` directly rather than the plugin harness — it takes the anonymize closure as an argument, so a failing one is enough and no fake anonymizer is needed.

The `[]any` inconsistency you spotted is settled the way you proposed: a bare string is anonymized like the text block next to it, anything else gets the notice.

### The three carried over

Opened #14 (placeholders never restored in response tool calls), #15 (OpenAI `tool_calls`), #16 (`thinking` blocks, with your point about narrowing `default:` to the known attachment types — including the caveat that a `signature` must survive byte for byte).

The PR body now names the first two as limits of this patch. I took your second option for the format question: the node documentation says which format is protected and mentions the loss of streaming on agentic turns, rather than the title implying both formats.

On the placeholder cap in `buildInstructionText` — I left it alone. It is a real concern and tool traffic does make the mapping much larger, but a cap changes what gets restored on the way back, and that deserves its own measurement rather than a number picked inside this PR. Say the word if you would rather see it here.

`go test ./plugins/pseudonymizer/` and `go test ./internal/...` pass.

Bornholm:
I re-read f44fb22 end to end. The pairing holds: nothing in the tool branch reaches `removedParts` any more, the notice keeps the block in place, and the ten tests exercise the branches I asked about, `remove`, `block`, mixed content, object content, bare string, anonymizer failure. What I requested is done.

I'm approving, with one question I want answered before merge and three follow-ups.

### Does the top-level `system` survive on the Messages route?

`extractMessagesJSON` only pulls the `messages` array. The plugin prepends its own `system` message to that array. `applyModifiedMessages` then rebuilds the messages with `ConvertAnthropicMessagesJSON`, whose doc says the `system` prompt "is expected to be carried separately". And `llm.WithMessages` replaces `opts.Messages` wholesale, including the `system` messages `ParseMessagesRequest` produced from the request body.

If I read this right, any node that rewrites messages on the Messages route drops the client's top-level `system`, and the provider only sees the pseudonymizer instruction. That predates your patch and lives in the adapter, not in the plugin. But your patch makes the rewrite happen on every agentic turn, so it would now be the normal case rather than the odd one.

I haven't run it. Could you confirm on a real session against f44fb22 that the provider still receives Claude Code's system prompt? If it doesn't, I'll open an issue and we fix it in the adapter. It shouldn't hold this PR, but I'd rather know before it ships.

### Follow-ups

1. Size bound on tool content. The `isToolPart` branch runs before `attachmentText`, and `attachmentText` is where `max_attachment_bytes` and `max_attachment_chars` apply. A `tool_result` carrying a whole file now goes to the NER pass unbounded, every turn. A truncation with a notice, like `attachmentTextPart` does, would close it. I can take it here if you'd rather not, otherwise I'll open a ticket.
2. Widen #16. `server_tool_use`, `web_search_tool_result`, `code_execution_tool_result` and the `mcp_tool_*` pair fall into `default:` the same way `thinking` does. #16 only names `thinking`. I'll add the others so the "narrow `default:` to known attachment types" fix covers them all at once.
3. One doc line on `verification_on_leak`. A verification leak on a tool block, with any policy other than `block`, sends the whole request through untouched via `passthroughOutput`. Same as the `text` path, and the event fires, so nothing is silent. But with tool traffic that's a whole file rather than a sentence, and I'd want that spelled out next to the paragraph you added.

The fail-open on anonymizer failure is fine by me. It mirrors the `text` branch and the `passthrough` event makes it visible.

scriptted:
Confirmed, and it does not survive. I ran it against f44fb22 with a control, because the two possible answers do not lead to the same fix.

**With a rewriting node** (`acme/gpt-4o-mini`, pseudonymizer), a top-level `system` on `/api/v1/messages`, this is the entire body the provider received:

```json
{"messages":[{"content":"IMPORTANT : certaines informations personnelles ou sensibles de cette conversation ont été remplacées par des jetons de substitution tels que ⟦PERSON_1_702cc7⟧. …","role":"system"},{"content":"Bonjour, je m'appelle ⟦PERSON_1_702cc7⟧.","role":"user"}],"model":"gpt-4o-mini","max_completion_tokens":256}
```

The client's system prompt is gone, and the only `system` the provider sees is the plugin's own instruction — exactly what you described.

**Without a rewriting node** (`acme/e2e-fast`, same route, same shape), the prompt arrives intact. So it is the rebuild performed when a node rewrites messages, not the Messages route on its own. That matches your reading of `ConvertAnthropicMessagesJSON` and `llm.WithMessages`, and it puts the fix in the adapter.

Worth stating plainly: for an agentic client every turn goes through that rebuild, so the provider never sees the client's system prompt at all — not the tool policy, not the output format, nothing. It is not a degradation, it is the instructions being dropped silently.

I did not push the test: it fails, and this PR is approved. It is yours to place — in the issue you open, or here with a `t.Skip` pointing at it, whichever you prefer.

<details>
<summary><code>test/e2e/messages_route_test.go</code> — the failing case and its control</summary>

```go
//go:build e2e

package e2e

import (
	"bytes"
	"encoding/json"
	"io"
	"net/http"
	"strings"
	"testing"
	"time"
)

// messagesRequest drives the Anthropic Messages route, where the system prompt
// is a TOP-LEVEL field rather than a message — which is the whole point of this
// file. `chatMessages` cannot be reused: it posts to the OpenAI route, where a
// system prompt is just another entry in `messages`.
func messagesRequest(t *testing.T, token, modelName, system string, messages []map[string]any) chatResult {
	t.Helper()

	body := map[string]any{
		"model":      modelName,
		"max_tokens": 256,
		"messages":   messages,
	}
	if system != "" {
		body["system"] = system
	}

	req, err := http.NewRequest(http.MethodPost, env.baseURL+"/api/v1/messages", bytes.NewReader(mustJSON(body)))
	if err != nil {
		t.Fatal(err)
	}
	req.Header.Set("Authorization", "Bearer "+token)
	req.Header.Set("Content-Type", "application/json")
	req.Header.Set("anthropic-version", "2023-06-01")

	resp, err := (&http.Client{Timeout: 3 * time.Minute}).Do(req)
	if err != nil {
		t.Fatalf("messages request: %v", err)
	}
	defer resp.Body.Close()
	raw, _ := io.ReadAll(resp.Body)

	res := chatResult{Status: resp.StatusCode, Body: string(raw)}
	var parsed struct {
		Content []struct {
			Text string `json:"text"`
		} `json:"content"`
	}
	if json.Unmarshal(raw, &parsed) == nil && len(parsed.Content) > 0 {
		res.Content = parsed.Content[0].Text
	}
	return res
}

func TestMessagesRoute_ClientSystemPromptSurvivesARewritingNode(t *testing.T) {
	before := len(env.provider.Requests())

	const systemPrompt = "Tu es l'assistant du client, et tu réponds toujours en français."
	res := messagesRequest(t, tokenAlice, modelTagStrategy, systemPrompt, []map[string]any{
		{"role": "user", "content": "Bonjour, je m'appelle Jean Dupont."},
	})
	if res.Status != 200 {
		t.Fatalf("status = %d, body = %s", res.Status, res.Body)
	}

	upstream := env.provider.RequestsSince(before)
	if len(upstream) != 1 {
		t.Fatalf("upstream calls = %d, want 1", len(upstream))
	}
	sent := upstream[0]

	// The node did rewrite: without this the test could pass on a request the
	// plugin never touched, and prove nothing about the rebuild.
	if strings.Contains(sent.Raw, "Jean Dupont") {
		t.Fatalf("the pseudonymizer did not rewrite this request: %s", sent.Raw)
	}

	if !strings.Contains(sent.Raw, systemPrompt) {
		t.Errorf("the client's system prompt never reached the provider.\nsent: %s", sent.Raw)
	}

	var carried bool
	for _, m := range sent.Messages {
		if m.Role != "system" {
			continue
		}
		if text, ok := m.Content.(string); ok && strings.Contains(text, systemPrompt) {
			carried = true
		}
	}
	if !carried {
		t.Errorf("the system prompt did not reach the provider as a system message: %+v", sent.Messages)
	}
}

// The control: same route, same shape, a model with no node rewriting anything.
func TestMessagesRoute_ClientSystemPromptWithoutARewritingNode(t *testing.T) {
	before := len(env.provider.Requests())

	const systemPrompt = "Tu es l'assistant du client, et tu réponds toujours en français."
	res := messagesRequest(t, tokenAlice, modelFast, systemPrompt, []map[string]any{
		{"role": "user", "content": "Bonjour."},
	})
	if res.Status != 200 {
		t.Fatalf("status = %d, body = %s", res.Status, res.Body)
	}

	upstream := env.provider.RequestsSince(before)
	if len(upstream) != 1 {
		t.Fatalf("upstream calls = %d, want 1", len(upstream))
	}
	if !strings.Contains(upstream[0].Raw, systemPrompt) {
		t.Errorf("the system prompt is lost even with no rewriting node.\nsent: %s", upstream[0].Raw)
	}
}
```

</details>

### Your three follow-ups

1. **Size bound on tool content** — mine to fix, since my patch is what routes tool payloads past `attachmentText` and its two limits. I'd rather not grow an approved PR: merge this one, and I'll follow with the truncation-plus-notice modelled on `attachmentTextPart`. Open the ticket or let me, as you prefer.
2. **Widen #16** — yours, thanks. `server_tool_use`, `web_search_tool_result`, `code_execution_tool_result` and the `mcp_tool_*` pair all land in `default:` for the same reason, and narrowing `default:` covers them in one go.
3. **`verification_on_leak`** — done in 9a44038, next to the two other limits. It says the passthrough applies to the whole request and that tool traffic carries a file where a message carries a sentence.

Bornholm:
Thanks for the control run. That settles it, and it puts the fix where you said: the adapter. I opened #17 with your test and the body you captured. We'll take it.

Your split of the follow-ups works for me. Open the ticket for the size bound when you're ready, it's yours. #16 is mine to widen. 9a44038 reads well.

One thing came up on the last pass that nobody had raised, and I want it on the record before this merges.

### Secrets in tool content are redacted without a mapping

`builtin_secret_patterns` is on by default. When go-anon matches a secret it skips every strategy, no mapping, no cache, and emits a marker that is irreversible by design (`anonymizer.go:166`, `placeholder.go:74` in v0.3.0). That was fine when the input was a human message. With this patch the input is what an agent reads.

So the scenario is: the agent reads a config file that holds an API key, the model sees the marker instead of the key, the agent rewrites the file with an `Edit` or `Write` call, and the marker lands on disk. The original value is gone. #14 won't help here, there is nothing to restore, and unlike a placeholder collision the loss is silent.

I'm not asking you to change the patch. We'll add the warning to the pseudonymizer doc: behind this node, an agent must not rewrite a file that contains a secret, and operators who run agents on repositories with committed credentials should turn `builtin_secret_patterns` off for that virtual model or accept the loss. If you have a better answer than a warning, say so, otherwise I'll write it.

Merging.

Bornholm (approved):
Approved, see the comment above. Follow-ups in #17, #16 and the size-bound ticket.

PR #26, opened 2026-09-12
OIDC authentication compatible with multi-tenancy

kvitrvn:
## What

Make OIDC authentication compatible with multi-tenancy:

  - derive the public base URL from each tenant’s hostname;
  - generate and cache OIDC providers per provider/hostname pair;
  - keep OAuth callbacks on the tenant’s hostname;
  - require an absolute base URL when multi-tenancy is enabled;
  - document the required OAuth and DNS configuration.

Closes #

## Checklist

- [x] Every commit is signed off (`git commit -s`, see [CONTRIBUTING.md](../CONTRIBUTING.md))
- [x] `go test ./...` passes
- [x] `make test-integration` passes, if stores were touched
- [x] `make generate` was run, if `.templ` files or the tailwind CSS were edited
- [x] Documentation updated where relevant (`docs/`)


Bornholm (changes requested):
The shape of this is right. Resolving the base URL in the outermost wrapper of `server.Run`, describing providers as factories rather than as instances, and stripping the host scope before persisting the identity: all three hold up, and the comment in `host_scoped_provider.go` explains why better than I would have. As a bonus, OIDC discovery is no longer fetched twice per provider at startup, which the description does not mention.

Three things before this merges, only one of them really blocking.

## The request host is not normalised

`newTenantBaseURLResolver` keeps the host exactly as the client sent it (`perHost.Host = r.Host`, base_url_resolver.go:37) while the `MatchesHost` guard goes through `slugFromHost`, which does `strings.ToLower(stripPort(host))` (middleware.go:104). Two families of Host therefore clear the guard without producing the right callback:

- mixed case, `ACME.Xolo.Example.Com`
- explicit port, `acme.xolo.example.com:443`

In both cases `oidcCallbackURL` builds a redirect_uri that does not match, byte for byte, the one declared at the IdP, and the login fails without telling the user anything. And since `HostScopedProviderName` derives its name from that host, every variant adds an entry to `p.registered` and to goth's global registry, for good.

Careful, `strings.ToLower(r.Host)` is not enough: it still lets the port variant through. The most robust fix is not to start from `r.Host` at all, but to rebuild the host from the resolved slug and the pattern suffix, keeping the port of the configured base URL. Failing that, normalise exactly as `slugFromHost` does and put the configured port back. A `mixed-case host` case and a `host:port` case in `TestNewTenantBaseURLResolver` would pin both down.

## gothic's `?provider=` wins over the context value

The comment on `withContextProvider` says that "gothic reads that context value before the route parameter". That is true of the route parameter, false of the query string. In goth v1.82.0, `getProviderName` reads in this order:

```go
if p := req.URL.Query().Get("provider"); p != "" { return p, nil }
if p := req.URL.Query().Get(":provider"); p != "" { return p, nil }
// ... and only then
if p, ok := req.Context().Value("provider").(string); ok { return p, nil }
```

So `/auth/oidc/providers/github?provider=github@other-tenant.example.com` starts a `BeginAuth` on another host's provider, from any host. The flow does not complete, the gothic cookie being host-only, but the invariant this PR installs ("the host picks the provider") can be bypassed by any client, and the defence rests on a property of the session store that nothing states anywhere. Dropping the `provider` and `:provider` keys from `r.URL.Query()` in `withContextProvider` settles it in two lines.

## The registry and discovery, taken together

These two are harmless on their own, less so combined. For every (provider, host) pair never seen before, `Resolve` builds the provider outside the lock, and on the fallback path `openidConnect.NewNamed` fetches the discovery document with `http.DefaultClient`: no context, no timeout. So an unauthenticated `GET /auth/oidc/providers/x` with a forged Host triggers an outbound request and a memory entry that is never released. The per-IP rate limiter on `/auth/oidc/` is the only brake, and the normalisation above removes most of the leverage. Still, I would put an `http.Client` with an explicit timeout on that fallback, and maybe a bound on the registry size.

On evicting registry entries, I expect nothing in this PR: one provider per tenant ever seen is small, and invalidation depends on the provisioning API. Worth keeping in mind for instances with high turnover, along with the client secret that stays in memory after a tenant is deleted.

## The rest, your call

In multi-tenant mode no factory runs at startup, whereas the single-tenant branch fails the boot on a misconfigured provider. A broken provider therefore gives a green start, then a 404 for the first user of each tenant. Building a throwaway instance per provider at startup, only to validate it, without registering it in goth, would give the feedback loop back.

And that 404 is a plain-text `http.NotFound` for every `resolveProvider` error, unknown provider and unreachable IdP alike. An "unknown provider" sentinel answering 404 and a 502 rendered through `common.HandleError` for the rest would tell the two apart.

One last detail: `oidcCallbackURL` concatenates with `fmt.Sprintf`, so `XOLO_HTTP_BASE_URL=https://xolo.example.com/` yields a double slash in the callback. The defect predates this PR, but the PR now requires an absolute URL in multi-tenant mode, a value people readily type with a trailing slash. A `strings.TrimSuffix` in `oidcCallbackURL` or a refusal in `validateMultitenantBaseURL`, whichever you prefer.

Bornholm (approved):
All three items from my previous review are fixed, and fixed the right way. Lifting the block.

`CanonicalHost` (middleware.go:97) pulls the slug out with the same validation `Resolve` uses and rebuilds the host from `hostPrefix + slug + hostSuffix`, and the resolver puts the port back from the configured base URL through `net.JoinHostPort`. That kills both the mixed-case and the `host:port` variant, rather than only the one a `ToLower` would have caught.

`withContextProvider` now deletes `provider` and `:provider` from the query before setting the context value, on an `r.Clone`, so the original request is left alone. I like that `handler_provider_test.go` calls the real `gothic.GetProviderName` instead of restating the lookup order: if goth ever reorders it, the test says so.

Discovery is read once at startup behind a 10s client and the `NewNamed` fallback is gone, so resolving a tenant does no network I/O at all. The registry-plus-discovery pairing I was worried about no longer exists: a forged host takes a 404 well before it reaches `Resolve`.

Two things I would still like, neither of them blocking.

## `jwks_uri` should not be fatal

`validateOIDCDiscovery` marks it `required: true` and the error travels up to fail the boot, in single-tenant mode too. Before this PR, a document without `jwks_uri` left interactive login perfectly usable and only turned off API token validation. That field is used by the `oidctoken` and `oauth2token` path alone, so requiring it to start means refusing to boot over something the login flow never reads. An instance sitting in front of a non-conformant IdP (opaque tokens plus introspection) stops starting after an upgrade, and nothing in the description says so.

I would keep `issuer`, `authorization_endpoint` and `token_endpoint` required, and downgrade a missing `jwks_uri` to a warning that names what it disables. It also lines the policy up with the Gitea branch, which still swallows its discovery errors silently. If you would rather keep it strict, then it belongs in the description and the changelog as a breaking change.

## Normalise the base URL once

`validateMultitenantBaseURL` parses `strings.TrimSpace(baseURL)` but the untrimmed `c.HTTP.BaseURL` is what reaches `http.WithBaseURL`, `newTenantBaseURLResolver` and `oidcCallbackURL`. A stray space or newline (easy to get from a `.env` or a ConfigMap) clears validation, then fails further along, or in single-tenant mode where the check does not run, quietly ends up in every generated URL, OAuth redirect_uri included. Trimming `c.HTTP.BaseURL` at config parse time, so the validated value is exactly the one used afterwards, closes it.

## Notes, nothing to do

In single-tenant mode `CanonicalHost` returns `(strings.ToLower(stripPort(host)), true)` with no validation, so an empty or forged host comes back through. It is unreachable today, the resolver only being installed when multi-tenancy is on, but it does not match the contract the doc comment above it states. Returning false on an empty host would keep the two in agreement.

On the throwaway instances `buildStartupOIDCProviders` builds: keep them. Discovery being validated just above makes them redundant for the current factories, but they are the only thing that would catch a future factory failing for another reason, and they cost one startup pass.

Registry eviction is still open and still fine by me. Worth a follow-up issue rather than anything here: a deleted tenant keeps its goth entry, and its client secret, until the process restarts.

PR #49, opened 2026-09-15
fix(oidc): oidc login failures caused by oversized provider session cookies

### LINKED ISSUE #50: OIDC callback fails when the authorized provider session exceeds the cookie size limit
### What happened?

hen Xolo is configured with Keycloak as an OIDC provider, authentication succeeds at Keycloak, but Xolo fails
  while processing the callback and redirects the user back to the login page.

  Xolo logs the following error:

  could not complete user auth
  error="securecookie: the value is too long"

  The affected user is already provisioned in Xolo with the correct provider and subject, so the failure occurs
  before user authorization.

  The underlying issue appears to be that gothic.CompleteUserAuth persists the authorized provider session after
  exchanging the authorization code. This session contains the access token, ID token, and refresh token. With
  Keycloak, the serialized session can exceed the browser’s cookie size limit.

  Expected behavior:

  - Xolo should validate the callback state.
  - Xolo should exchange the authorization code.
  - Xolo should fetch the authenticated user and create its local session.
  - Provider tokens should not be persisted back into the browser cookie when they are only needed during the
    callback request.

  Actual behavior:

  - The provider session exceeds the secure-cookie limit.
  - The callback fails.
  - Xolo redirects to /auth/oidc/logout and then back to the login page.

  Avoiding this persistence would also reduce unnecessary exposure of provider tokens to the browser.

### Steps to reproduce

1. Configure Xolo with Keycloak as a confidential OIDC provider.
  2. Configure Xolo to use its cookie-based session store.
  3. Enable multitenancy and provision a user with:
      - provider=keycloak
      - the Keycloak user’s realm sub

  4. Open the tenant’s Xolo URL.
  5. Select the Keycloak identity provider.
  6. Authenticate at Keycloak, or reuse an existing Keycloak SSO session.
  7. Observe that Keycloak successfully redirects to Xolo’s callback.
  8. Observe that Xolo redirects back to its login page.
  9. Check the Xolo logs and find securecookie: the value is too long.

### Xolo version

e2ac00bff8d3ca74e8ef5c2096f958c393bc25f3

### Environment

Deployment: Docker Compose
  Storage backend: PostgreSQL
  Identity provider: Keycloak 26.4
  Authentication: OIDC authorization code flow
  Browser: Firefox / Chromium

Relevant configuration:

XOLO_MULTITENANCY_ENABLED=true
  XOLO_MULTITENANCY_HOST_PATTERN={tenant}.xolo.localhost:3002
  XOLO_HTTP_AUTHN_AUTO_CREATE_USERS=false
  XOLO_HTTP_AUTHN_OIDC_PROVIDERS=keycloak
  XOLO_HTTP_AUTHN_OIDC_PROVIDER_KEYCLOAK_SCOPES=openid,profile,email
  XOLO_HTTP_SESSION_COOKIE_SECURE=false

  Client secrets and session keys have been omitted.

### Relevant logs

```text
level=ERROR
  msg="could not complete user auth"
  error="securecookie: the value is too long"

  request.method=GET
  request.host=sso-e2e.xolo.localhost:3002
  request.path=/auth/oidc/providers/keycloak/callback
  response.status=307

  The callback is immediately followed by:

  GET /auth/oidc/logout
  GET /auth/oidc/login

  No user lookup occurs before this failure.
```

kvitrvn:
## What

This change prevents Xolo from persisting the authorized OIDC provider session in the browser cookie after the
  authorization code exchange.

  ### Problem

  Goth stores the complete authorized provider session, including access, ID, and refresh tokens, in a cookie.
  Keycloak sessions can exceed the browser’s 4 KB cookie limit, causing the callback to fail with:

  securecookie: the value is too long

  Users were consequently redirected back to the login page even though authentication with Keycloak succeeded.

  ### Solution

  - Keep the initial OIDC state in the cookie until the callback.
  - Preserve state-token validation.
  - Exchange the authorization code and fetch the user within the callback request.
  - Do not persist the resulting tokens back to the browser.
  - Clear the temporary provider session after authentication.

  This also reduces unnecessary exposure of provider tokens to the browser.

  ### Verification

  - Added a regression test ensuring authorized provider sessions are not persisted.
  - Added coverage for OIDC state mismatch rejection.
  - Ran the complete Xolo test suite successfully.
  - Built and started the Docker image successfully.
  - Verified Chat-to-Xolo Keycloak SSO end to end without a second credential prompt.

Closes #50 

## Checklist

- [ x] Every commit is signed off (`git commit -s`, see [CONTRIBUTING.md](../CONTRIBUTING.md))
- [ ] `go test ./...` passes
- [ ] `make test-integration` passes, if stores were touched
- [ ] `make generate` was run, if `.templ` files or the tailwind CSS were edited
- [ ] Documentation updated where relevant (`docs/`)


Bornholm (changes requested):
The fix is sound. `complete_user_auth.go` is a faithful fork of `gothic.CompleteUserAuth` from goth v1.82.0, diffed line by line against the module cache, with one behavioural change: the `StoreInSession(providerName, sess.Marshal(), res, req)` call after `sess.Authorize` is gone, which is exactly the call raising "securecookie: the value is too long" on Keycloak sessions. State validation including the POST form branch, the deferred `gothic.Logout` and the `FetchUser` -> `Authorize` -> `FetchUser` sequence are all preserved, and the stricter `ParseForm` error handling is an improvement.

Two things I checked before trusting the change: `gothic.Store` and the Xolo session store share the same `CookieStore` but use different session names (`_gothic_session` against `xolo_auth_oidc`, `oidc/options.go:35`), so the deferred `Logout` cannot clobber the local session `storeSessionUser` writes afterwards; and nothing under `internal/` reads the authorized provider session after the callback, so dropping persistence leaves no consumer behind. The package tests pass.

Three points before this comes out of draft.

## The login-start route still persists

Only the callback was moved to the local `completeUserAuth`. `internal/http/middleware/authn/oidc/provider.go:17` still calls `gothic.CompleteUserAuth`, whose `Authorize` path runs the `StoreInSession` this PR removes.

Reaching it takes a request carrying both a pending `_gothic_session` cookie and a valid code and state, which the IdP redirect never produces since it targets `/callback`, and the securecookie error would fall through to `BeginAuthHandler` rather than reach the user. So this does not bring the reported failure back for real users. But the oversized-session persistence path, and the token exposure the PR sets out to remove, still live in the codebase on a route a hand-crafted or replayed URL can reach. Either route `handleProvider` through `completeUserAuth` too, or say in a comment why the login route keeps upstream behaviour on purpose.

## The regression test asserts through a panic, and never looks at the cookies

`oversizedSession.Marshal` panics when the session is authorized (`complete_user_auth_test.go:104-107`). `testing` does not recover a panic raised in test code, so a regression aborts the whole oidc test binary instead of failing one case: the other tests never report, the failure reads as a stack trace instead of an assertion, and the `t.Cleanup` that restores `gothic.Store` and the goth provider registry stops being reliable. The guard is indirect too, since a persistence path that does not call `Marshal` would pass in silence.

The bigger gap is the recorder. `completeUserAuth` gets a throwaway `httptest.NewRecorder()` at line 44 that nothing inspects, so the test never asserts what #50 is about: that no oversized `_gothic_session` cookie is written back, and that the only `Set-Cookie` left is the `MaxAge=-1` deletion from the deferred `Logout`.

Record the call in a field and assert it with `t.Errorf` after `completeUserAuth` returns, keep a named recorder and assert its `Set-Cookie` headers, and add a mismatched-state case going through `completeUserAuth` rather than calling `validateState` directly.

## Pin the fork to the version it mirrors

The doc comment says the function mirrors `gothic.CompleteUserAuth` but names neither the version nor the intended delta. On the next goth upgrade, an upstream fix to state validation or token handling will not propagate here and nothing will signal that a re-diff is due. Naming v1.82.0 and the removed call in the comment is the cheap version. Is there a plan beyond that, an upstream issue asking for a no-persist variant, or a diff step in CI?

Minor: the new test imports `golang.org/x/oauth2` directly while `go.mod:124` still marks it `// indirect`. Harmless today, since CI runs no tidy check, but the next `go mod tidy` will produce an unrelated diff.

PR #51, opened 2026-09-16
fix(oidc): apply validateOIDCDiscovery to Gitea and tolerate missing jwks_uri

### LINKED ISSUE #27: oidc: a discovery document without jwks_uri now fails startup, though only API token validation needs it
Raised while reviewing #26, not blocking it.

`validateOIDCDiscovery` (`internal/setup/authn_oidc_handler.go`) marks `jwks_uri` as required, and the error travels up through `buildOIDCProvider` to `getOIDCAuthnHandlerFromConfig`, which fails the boot. Single-tenant instances included.

```go
{name: "issuer", value: discovery.Issuer, required: true},
{name: "authorization_endpoint", value: discovery.AuthURL, required: true},
{name: "token_endpoint", value: discovery.TokenURL, required: true},
{name: "jwks_uri", value: discovery.JWKSURI, required: true},
```

Before #26, a discovery document without `jwks_uri` left interactive login untouched: `withJWKS` stayed nil and only API token validation was silently disabled.

```go
if err == nil && discovery != nil && discovery.JWKSURI != "" {
```

`jwks_uri` is read by the `oidctoken` and `oauth2token` authenticators alone, so requiring it to start means refusing to boot over a field the login flow never touches. The OIDC discovery spec does mark it required, so conformant providers are unaffected, but an instance sitting in front of a non-conformant IdP (opaque tokens plus introspection) stops starting after an upgrade, with no mention of it in the release notes.

The Gitea branch of the same function still swallows its discovery errors silently, so the policy is not uniform either.

## What to do

Keep `issuer`, `authorization_endpoint` and `token_endpoint` required. Downgrade a missing `jwks_uri` to a startup warning that names what it disables, so the instance boots with interactive login working and API token validation off for that provider.

If strictness is the deliberate choice instead, say so in the changelog and in `docs/fr/administration/installation/configuration.md`, and apply it to Gitea too.

Padiwa:
Closes #27.

## Problem

`validateOIDCDiscovery` marked `jwks_uri` as required in the OIDC discovery document. A non-conformant IdP — one that issues opaque tokens validated via introspection, or simply does not publish a JWKS — would therefore refuse to boot Xolo after the upgrade in #26, even though the interactive login flow never reads `jwks_uri` (only the `oidctoken` and `oauth2token` API-token authenticators do). Issue #27 is exactly this IdP-with-opaque-tokens shape.

The Gitea branch of the same function had no `jwks_uri` enforcement but silently disabled API-token validation when `jwks_uri` was absent, on the same registry as the named OIDC providers. The policy was therefore not uniform between named OIDC providers and the built-in Gitea provider.

### `validateOIDCDiscovery`

`jwks_uri` is no longer required. `issuer`, `authorization_endpoint` and `token_endpoint` remain mandatory; every other present endpoint, including `jwks_uri`, must still be an absolute `http(s)` URL when it is set. URL schemes are normalised to lowercase for the comparison so an IdP publishing `HTTPS://…` is accepted (RFC 3986 case-insensitivity).

### `buildOIDCProvider` (named OIDC providers)

When `jwks_uri` is missing, malformed, or present but not an absolute `http(s)` URL, a new helper `tolerableJWKSURI` emits one `slog.WarnContext` record (level `WARN`) named after the affected provider, attaches the discovered `introspection_endpoint` and `userinfo_endpoint` as `slog` attributes when present, and leaves `JWKSURL` empty on the descriptor. The provider stays in the registry with an empty `JWKSURL`:

- `oidctoken.validateToken` already short-circuits on `provider.JWKSURL == ""` (returns `errInvalidToken`, which the authentication loop continues on), so JWT id-token validation is silently disabled for this provider.
- `oauth2token.ProvidersForTokenValidation` filters on `IntrospectionURL` / `UserInfoURL` directly, not on `JWKSURL`, so introspection / UserInfo keep working when the discovery document exposes either endpoint. This is exactly the case issue #27 opens.

### `buildGiteaProvider` (new helper)

The Gitea configuration used to be an inline block inside `getOIDCAuthnHandlerFromConfig`. It is now a `buildGiteaProvider` function mirroring `buildOIDCProvider`, so the two stay symmetric and testable.

`XOLO_HTTP_AUTHN_PROVIDERS_GITEA_DISCOVERY_URL` is **optional**: when it is empty, the operator gets a startup warning, the discovery step is skipped, and the interactive-login factory is built from the static `AUTH_URL` / `TOKEN_URL` configuration (mirroring `buildOIDCProvider`'s discover-then-fallback rule: when both are present, the discovered values always win; the static config only matters on the empty-DiscoveryURL path, which is explicitly validated above). When `PROFILE_URL` is empty on the discovery path, the function falls back to `discovery.UserInfoEndpoint`.

When `DiscoveryURL` is configured, the document is fetched once and validated against `validateOIDCDiscovery`. A non-success status, a malformed body, or a missing / non-absolute `issuer`, `authorization_endpoint` or `token_endpoint` fails boot with a clear `could not configure gitea provider` error that names the affected slot. URL schemes are normalised to lowercase.

## Hardening

This PR additionally tightens one Gitea startup invariant: the previous code accepted a Gitea provider configured with only `KEY` and `SECRET` (no `DISCOVERY_URL`, `AUTH_URL`, or `TOKEN_URL`) and let the failure surface later as an opaque goth redirect error on the first login. Such a configuration was never functional, but it is now refused at boot with a clear `gitea provider requires either DISCOVERY_URL or both AUTH_URL and TOKEN_URL` message. Operators hitting this refusal just need to set either the discovery URL or the static OAuth endpoints.

The empty-DiscoveryURL path also validates `AUTH_URL` / `TOKEN_URL` (and `PROFILE_URL` when present) against the same absolute-http(s) shape as `validateOIDCDiscovery`, so a malformed entry fails boot in the same way as a malformed discovery field.

## Tests

- `TestBuildOIDCProviderToleratesMissingJWKSURI` — asserts the descriptor stays populated with `JWKSURL: ""` and that `IntrospectionURL` / `UserInfoURL` / `ClientID` / `ClientSecret` survive (issue #27 case). Asserts one warning with `provider=` and `discovery_url=` attributes and the optional endpoint attrs.
- `TestBuildOIDCProviderToleratesUnusableJWKSURI` — covers relative and `file://` schemes: same boot-succeeds behaviour, separate "not absolute http(s)" warning.
- `TestTolerableJWKSURIAcceptsUppercaseScheme` — asserts an `HTTPS://…` jwks_uri is accepted (RFC 3986 case-insensitive scheme).
- `TestTolerableJWKSURIDoesNotPanicOnUnparseableInput` — asserts a control character in jwks_uri is tolerated instead of crashing the dereference of `parsed.Scheme`.
- `TestBuildOIDCProviderRejectsInvalidDiscovery` — covers all three remaining mandatory endpoints (`issuer`, `authorization_endpoint`, `token_endpoint`) for both the missing and the relative-URL cases.
- `TestBuildGiteaProviderToleratesMissingJWKSURI` — five subtests: missing `jwks_uri` keeps the provider for introspection; `jwks_uri` present populates the registry; empty `DiscoveryURL` boots with the static config + warning; unusable `jwks_uri` is tolerated.
- `TestBuildGiteaProviderRejectsInvalidDiscovery` — mirror of the OIDC test: non-success status, malformed body, and missing / relative `issuer`, `authorization_endpoint`, `token_endpoint` all fail boot.
- `TestBuildGiteaProviderRequiresDiscoveryOrStaticAuth` — covers the new `gitea provider requires either DISCOVERY_URL or both AUTH_URL and TOKEN_URL` refusal.

The slog-capturing test helper (`captureHandler` + `withCapturedLogs`) replaces the previous `TextHandler` / `slog.SetDefault` machinery so assertions inspect typed attributes instead of rendered strings.

## Documentation

`docs/fr/administration/installation/configuration.md`:

- The auth-provider table lists `_DISCOVERY_URL` as a Gitea field and states the field is optional.
- A new "Fournisseur Gitea" subsection describes the empty-`DISCOVERY_URL` startup warning, the static-config fallback, and the discover-then-fallback rule on `authorization_endpoint` and `token_endpoint`: discovered endpoints always win over the static config because `validateOIDCDiscovery` rejects any document that omits them.
- The introspection paragraph is rewritten: a missing or unusable `jwks_uri` disables JWT id-token validation; `oauth2token` still validates opaque tokens via the discovery document's `introspection_endpoint` (or `userinfo_endpoint` as fallback).

## Operator-visible change

A new `slog.WarnContext` line at startup per affected provider, of one of these forms:

```
WARN oidc provider discovery document has no jwks_uri; JWT id-token validation is disabled for this provider provider=<id> discovery_url=<url> [introspection_endpoint=<url>] [userinfo_endpoint=<url>]
WARN oidc provider discovery document jwks_uri is not an absolute http(s) url; JWT id-token validation is disabled for this provider provider=<id> discovery_url=<url> jwks_uri=<raw value> [introspection_endpoint=<url>] [userinfo_endpoint=<url>]
WARN oidc provider discovery document jwks_uri is not a parseable url; JWT id-token validation is disabled for this provider provider=<id> discovery_url=<url> jwks_uri=<raw value> parse_error="<url.Parse error>"
```

The instance boots; interactive login works; JWT-signed id-tokens from this provider are not validated. If the discovery document publishes `introspection_endpoint` (or `userinfo_endpoint` as a fallback) the same provider still validates opaque API tokens through `oauth2token`.

For the Gitea provider, an extra `slog.WarnContext` is emitted on startup when `XOLO_HTTP_AUTHN_PROVIDERS_GITEA_DISCOVERY_URL` is empty, of the form:

```
WARN gitea provider has no discovery url; JWT id-token validation is disabled and interactive login falls back to the static AUTH_URL / TOKEN_URL configuration provider=gitea
```


Bornholm (changes requested):
Reviewed `3ec2730`. The PR does what issue #27 asks: `validateOIDCDiscovery` no longer requires `jwks_uri` while still checking it is absolute when present, both provider paths start with a warning instead of refusing to boot, and extracting `buildGiteaProvider` makes the two branches symmetric and testable. The added tests cover the three endpoints that stay mandatory on both sides, the warning, and the JWKS registry. That part is solid.

Three things to settle before merge, one of which lands on the exact use case the issue describes.

## Dropping the provider from providersWithJWKS also disables introspection

`internal/setup/authn_oidc_handler.go:331-352`, and the same shape in `buildGiteaProvider` at 246-263.

When `jwks_uri` is missing, `withJWKS` returns `nil` and the provider leaves `providersWithJWKS` entirely. That slice also feeds `Handler.ProvidersForTokenValidation()` (`internal/http/middleware/authn/oidc/handler.go:92`), which builds the `oauth2token` providers from `IntrospectionURL` and `UserInfoURL`. Neither field needs a JWKS.

So an IdP that publishes `introspection_endpoint` but no `jwks_uri`, which is the case issue #27 opens with ("an IdP issuing opaque tokens validated by introspection"), now boots but loses all API token validation, introspection included. The PR removes the startup refusal without restoring the capability the motivating example needed.

`oidctoken` already handles an empty JWKS URL cleanly: `validateToken` returns `errInvalidToken` when `provider.JWKSURL == ""` (`oidctoken/authenticator.go:119`), and the authentication loop at line 87 does `continue` on that error. Registering the provider with `JWKSURL: ""` and keeping `IntrospectionURL`, `UserInfoURL`, `ClientID` and `ClientSecret` would leave `oauth2token` working and cost nothing on the `oidctoken` side. The warning message and the matching doc paragraph would then say that JWT ID token validation is disabled, not token validation as a whole. If you prefer to keep the current behaviour, the two provider lists should be decoupled instead.

Was the full exclusion a deliberate call here, given that `oidctoken` already skips an empty `JWKSURL` on its own?

## Making GITEA_DISCOVERY_URL mandatory adds a startup refusal to the PR that removes one

`internal/setup/authn_oidc_handler.go:219-234`. `buildGiteaProvider` fails when `DiscoveryURL` is empty, and the error propagates through `getOIDCAuthnHandlerFromConfig` (line 121) up to the boot refusal. Two consequences.

Any Gitea instance configured with only `AUTH_URL`, `TOKEN_URL` and `PROFILE_URL` stops booting on upgrade. That was a valid configuration until now, and it is what `.env.dist` carried before this PR. And Xolo's startup now depends on the Gitea server being reachable with `/.well-known/openid-configuration` enabled, while the interactive login flow reads nothing from the discovery document.

Issue #27 does allow strict uniformity if it is documented, which the PR does (the "Breaking change" section plus the "Fournisseur Gitea" doc block). But what the issue actually criticised was Gitea silently swallowing its discovery errors, not the absence of a `DiscoveryURL`. This goes past the minimum, so I would like an explicit call on it rather than merging it as a side effect.

A non-breaking alternative: keep `DiscoveryURL` optional for Gitea, where empty means a startup warning and no JWKS registration while interactive login keeps running on the static URLs, and apply `validateOIDCDiscovery` strictly as soon as the URL is set. Same uniform policy, no broken deployments, no coupling between boot and Gitea. If you keep the break, it belongs in the release notes, not only in the French documentation.

## The Gitea login factory uses config URLs that nothing validates

`internal/setup/authn_oidc_handler.go:236-243`. `buildGiteaProvider` now requires a discovery document and validates its `authorization_endpoint` and `token_endpoint`, but the interactive login factory is still built from `gp.AuthURL`, `gp.TokenURL` and `gp.ProfileURL`, unlike `buildOIDCProvider`, which reuses the discovered endpoints (see `TestBuildOIDCProviderUsesDiscoveredEndpoints`).

Two effects. If the discovery document and the static config diverge, login silently uses the stale URLs. And nothing requires `AUTH_URL`, `TOKEN_URL` or `PROFILE_URL` any more, since the only startup gate is `DiscoveryURL`: an operator following the new documentation ("DISCOVERY_URL is mandatory") can configure Gitea with the discovery URL alone and get a goth provider with empty endpoints that only fails on the first login. Those fields have no `envDefault` (`internal/config/http.go:102-108`) and `validateOIDCDiscovery` never looks at the config.

Either build the Gitea factory from the discovered endpoints with the config URLs as fallback, mirroring `buildOIDCProvider`, or fail or warn at startup when `AuthURL` or `TokenURL` is empty or diverges from the document.

## Smaller points

`authn_oidc_handler.go:119-123`: Gitea errors come back as `errors.WithStack(err)` while named OIDC providers get `errors.Wrapf(err, "could not configure oidc provider %q", np.ID)`. An operator whose boot fails on `gitea provider discovery url is required` gets no hint about which provider slot or which environment variable is at fault. Worth a `errors.Wrap(err, "could not configure gitea provider")`, in a PR that is otherwise about making the two branches symmetric.

`authn_oidc_handler.go:207-210`: the comment offers an escape hatch the code does not have, saying operators can configure equivalent `AUTH_URL` / `TOKEN_URL` and a self-hosted `/.well-known/` path. An empty `DiscoveryURL` fails unconditionally at line 224 and the static URLs are never consulted for validation. State the single requirement instead.

`docs/fr/administration/installation/configuration.md:48-50`: the new sentence tells operators to set `XOLO_HTTP_AUTHN_OAUTH2TOKEN_ENABLED=true` to validate opaque tokens "on an IdP that has a `jwks_uri`". An opaque-token IdP is precisely the one without a `jwks_uri`, so the advice does not cover the case it claims to. This follows from the first point and should move with whatever you decide there.

`internal/setup/authn_oidc_handler.go:420-450`: `jwks_uri` is no longer required, but the loop still applies the absolute-http(s) check to any non-empty value, so an IdP publishing a relative `jwks_uri`, the same family of non-conformance #27 targets, still breaks the boot. It is documented and deliberate, but treating an unusable `jwks_uri` like a missing one (warning plus exclusion from the registry) would make the tolerance consistent.

`internal/setup/authn_oidc_handler_test.go:251-330`: the warning assertions swap `slog.SetDefault` for a `TextHandler` writing to a buffer, then match rendered attributes (`provider=opaque-idp`, `discovery_url=`). That breaks if the default handler format changes, and `slog.Default` is process-global, so any test in the package moving to `t.Parallel()` turns it into a data race. The "exactly once" count also relies on nothing else logging during the test. A test `slog.Handler` capturing structured attributes would be sturdier. Cleanup via `t.Cleanup` is correctly done.

Finally, cosmetic: the documentation uses `https://gitea.example.com/.well-known/openid-configuration` while `.env.dist:31` uses `gitea-server`, which the other Gitea variables in that file already use.

Bornholm (changes requested):
Reviewed at `30e6398`.

The three blocking points from the previous round are closed, and I checked each one in the tree. The JWKS descriptor now survives a missing `jwks_uri`: it is kept with an empty `JWKSURL` and an intact `IntrospectionURL` / `UserInfoURL` / `ClientID`, so `oauth2token` keeps validating opaque tokens through introspection while `oidctoken` short-circuits on the empty URL. That is the case issue #27 is about. `GITEA_DISCOVERY_URL` is optional again, with a warning and a static fallback that `requireAbsoluteHTTPURL` now validates. And the discovery path feeds `discovery.AuthURL` / `discovery.TokenURL` to the factory instead of the static config.

Four things I would like changed. The first two are the reason I am requesting changes; the last two are cheap enough to fold in while you are here.

### PROFILE_URL escapes validation on the discovery path

`internal/setup/authn_oidc_handler.go:275-277` only rewrites `profileURL` when it is empty. A non-empty malformed value, a relative path or a `file://` URL, goes straight into `gitea.NewCustomisedURL`. The static branch twenty lines below (`:291-295`) validates the exact same field with `requireAbsoluteHTTPURL`.

The failure then surfaces on the first profile fetch, as an opaque goth error. That is the deferred failure this PR removes everywhere else, so leaving it on one branch is the part that bothers me. Hoisting the check above the `discoveryURL != ""` branch covers both paths in one line, and a sub-test mirroring "empty DiscoveryURL with malformed PROFILE_URL fails boot" on the discovery path would pin it.

### DISCOVERY_URL is never checked before the fetch

On the discovery path, `AUTH_URL`, `TOKEN_URL` and `PROFILE_URL` each get an explicit `requireAbsoluteHTTPURL`, but the URL actually being fetched does not (`:249`). A value with no scheme travels through `http.NewRequestWithContext` and dies inside `client.Do` with `unsupported protocol scheme`, wrapped as `could not configure gitea provider`. The operator gets none of the clear message every other field produces.

Boot still fails, so nothing is broken. It is the diagnostic that is worse than the one next to it. `buildOIDCProvider` has the same gap, and both deserve the same call.

### The new boot refusal is not in the operator docs

A Gitea provider with only `KEY` and `SECRET` used to boot and fail at the first login. It now refuses to start (`:230-234`). The hardening is right, that configuration never worked. But `docs/fr/administration/installation/configuration.md` describes the empty `DISCOVERY_URL` path in terms of a warning and a static fallback, and the refusal only shows through indirectly. There is no CHANGELOG in the repo, so the error string is currently the only place this prerequisite is written down.

One sentence in the "Fournisseur Gitea" section saying that either `DISCOVERY_URL` or both `AUTH_URL` and `TOKEN_URL` are required, otherwise the server refuses to start.

### withCapturedLogs mutates slog.Default globally

`internal/setup/authn_oidc_handler_test.go:124-140` swaps the global default logger. The `t.Cleanup` restore is correct and `captureHandler` fixes the coupling to the rendering format from the last round. What is left is that the mutation is guarded by a comment only: the first `t.Parallel()` added to this package turns the capture into a data race, and the exact warning counts break silently the day another code path emits a record carrying the same substring.

Either assert "at least one" where only the presence of the warning matters, or pass an explicit `*slog.Logger` into the production paths that log here.

### Not asking for a change

With `DISCOVERY_URL` set, boot now depends on Gitea being reachable. That is the uniform policy issue #27 asked for, and the optional `DISCOVERY_URL` keeps a network-free startup available. Worth a line in the release notes for upgrades where Gitea might be down.

Bornholm (commented):
Third pass on `f3fec8a`. Everything asked in the two previous rounds is in the tree, so this one is short.

## Closed since last round

- The JWKS descriptor survives a missing `jwks_uri`: `JWKSURL` stays empty, `IntrospectionURL`, `UserInfoURL` and the client credentials survive. That is the #27 case.
- `GITEA_DISCOVERY_URL` is optional. The static path validates `AUTH_URL` and `TOKEN_URL` with `requireAbsoluteHTTPURL`, the discovery path validates `DISCOVERY_URL` before the fetch, and `PROFILE_URL` is checked on both paths when set.
- The logger is injected into `buildGiteaProvider` and `buildOIDCProvider`, so the tests no longer touch `slog.Default`.
- `buildOIDCProvider` never returns a nil descriptor without an error, so the dereference at line 149 is safe, and the conditional append on the Gitea side is right.

## Two small asks

### Empty `PROFILE_URL` on the static path still fails at first login

On the static path an empty `PROFILE_URL` passes boot. goth v1.82.0 builds the profile request as `p.profileURL + "?access_token=..."` (`providers/gitea/gitea.go:103`) with no branch for an empty URL, so the first login fails with an opaque `unsupported protocol scheme`. That is the deferred failure the Hardening section removes elsewhere, and the discovery path already refuses to boot in the symmetric case (no `PROFILE_URL`, no `userinfo_endpoint`). The comment above the check calls `PROFILE_URL` an optional UserInfo fallback for goth, which is not what the Gitea provider does.

Either require `PROFILE_URL` on the static path the way the discovery path does, or fix the comment and the documentation so an operator knows a static Gitea configuration without it breaks at the first login. I would take the first.

### `captureHandler` groups only the `WithAttrs` attributes

`Handle` prefixes `h.attrs` with `h.group` but leaves the record's own attributes flat, while the helper's comment says the group prefix is honoured. No current test combines `WithGroup` with per-call attributes, so nothing breaks today. Either group the record attributes too, or narrow the comment.

## Not asking for a change

- The description still says a present `jwks_uri` must be an absolute URL. The code, the docs and `TestBuildOIDCProviderToleratesUnusableJWKSURI` tolerate it with a warning, which is the behaviour agreed in round one. Only the sentence is stale; worth fixing before merge so the PR reads like the code.
- With `GITEA_DISCOVERY_URL` set, boot now depends on Gitea being reachable. Already noted in round two and covered in `configuration.md`; there is no changelog to carry it further.

Bornholm (commented):
## Conclave review

No changes requested. The PR delivers what issue #27 asks for:

- `validateOIDCDiscovery` no longer requires `jwks_uri`. The issuer, `authorization_endpoint`, and `token_endpoint` stay mandatory; other present endpoints must be absolute.
- `tolerableJWKSURI` warns and returns an empty string for a missing, unparseable, or non-absolute `jwks_uri`. The descriptor stays in `providersWithJWKS` with `IntrospectionURL` / `UserInfoURL` / `ClientID` intact, so opaque token validation still works.
- `GITEA_DISCOVERY_URL` is optional. `DISCOVERY_URL` / `AUTH_URL` / `TOKEN_URL` / `PROFILE_URL` all go through `requireAbsoluteHTTPURL` before any fetch. The discovery path feeds `discovery.AuthURL` and `discovery.TokenURL` to the goth factory.
- The logger is injected (no `slog.Default` mutation in tests). Gitea errors are wrapped with "could not configure gitea provider".
- `go test ./internal/setup/` passes.

### Notes

- `pi-reviewer-1` failed to produce a valid report, so only one reviewer contributed findings. No issues were raised by the successful reviewer.
- A few low-severity consistency points are open (in-code comment and documentation describe `PROFILE_URL` as optional on the static path, while boot is now refused without it), but none are blocking.

PR #52, opened 2026-09-16
fix(authn): let a pending invitation stand for pre-provisioning

scriptted:
## What

With `XOLO_HTTP_AUTHN_AUTO_CREATE_USERS=false`, the invitation system cannot work at all.

The bridge middleware runs before every route, `/join/{token}` included. An invitee who has no account yet is answered 403 — *"Aucun compte Xolo n'est associé à cette identité"* — before ever reaching the page that would accept their invitation. So the two settings are mutually exclusive: an administrator who sends an invitation on such an instance has to provision the account by hand, which means knowing the identity provider's subject in advance.

This PR treats a pending **targeted** invitation as what it already is — pre-provisioning. An administrator named that address on purpose, so it joins the existing `DefaultAdmins` and application exceptions.

### The invitation grants the account, not its activation

`ActiveByDefault` keeps deciding that, as it does for any other identity. An administrator who turned it off wants to vet accounts, and an invitation is no reason to bypass the gate they chose.

Nothing is lost by waiting: `/join/{token}` is the one route that does not assert `authz.Active()`, so an invitee whose account is still inactive accepts the invitation and gets the membership and role right away. Only the rest of the instance waits for an administrator.

### A deliberately narrow exception

- **targeted invitations only** — an open link names nobody and would become a back door around `AutoCreateUsers`;
- **still-acceptable invitations only**, uses count included (the store filters on revocation and expiry, `IsInviteValid` covers the rest);
- **within the request tenant only**, resolved through the invitation's organization.

A lookup failure is never fatal — the identity simply falls back to the configured provisioning policy.

### Case-insensitive invitee matching

Same bug class, found on the way: invitee addresses were compared exactly, in the store lookup and in the join handler. The case an administrator types and the one an identity provider returns rarely agree, and a mismatch used to hide the invitation from its own addressee — silently, in `/no-org` and in the profile page as much as in the join flow.

### Tests

- `internal/http/middleware/bridge` — invited identity accepted, its activation left to `ActiveByDefault`, case-insensitive match, and three refusals that must keep working: another tenant's invitation, a revoked one, an untargeted one.
- `internal/adapter/gorm` — `ListPendingInvitesForEmail` through `eachBackend()`, asserting the case-insensitive lookup, the expiry/revocation filters and the organization preload on SQLite and PostgreSQL.

## Checklist

- [x] Every commit is signed off (`git commit -s`, see [CONTRIBUTING.md](../CONTRIBUTING.md))
- [x] `go test ./...` passes
- [x] `make test-integration` passes, if stores were touched
- [x] `make generate` was run, if `.templ` files or the tailwind CSS were edited — n/a, no template touched
- [x] Documentation updated where relevant (`docs/`)


Bornholm (changes requested):
The fix holds up. `hasPendingInvite` only grants the exemption to a targeted invitation that is still valid (`IsInviteValid` catches the exhausted-uses case the SQL query does not filter) and belongs to an organization of the request tenant, and a lookup error falls back to the configured policy. I checked the two claims the description rests on: `/join/` is indeed the only route mounted outside the `isActive` wrapper, and activation stays governed by `ActiveByDefault`, which defaults to `false`. The three refusals that matter are tested.

Nothing below blocks the intent of the change, but a few of them should be settled before this lands.

## Tenant scoping stops at the bridge

`ListPendingInvitesForEmail` takes no `TenantID`, and its two other callers (`internal/http/handler/webui/no_org.go:31` and the profile handler) show a user every pending invitation sent to their address, across tenants. `join.acceptInvite` fetches the invitation by ID alone and never compares `invite.Org().TenantID()` to the request tenant before `AddMember`. In multi-tenant mode a user of tenant A can accept an invitation issued by an organization of tenant B and get a membership and a role outside their tenant.

This predates the PR, but the documentation added here now claims a scoping the acceptance path does not enforce. Either add a `model.TenantID` parameter to `port.InviteStore.ListPendingInvitesForEmail` and assert the tenant in `getJoinPage` and `acceptInvite`, the way `hasPendingInvite` already does, or narrow the doc sentence to the pre-provisioning exemption alone.

## `DefaultAdmins` still compares case-sensitively

Three lines above the case fix, `internal/http/middleware/bridge/middleware.go:69` does `slices.Contains(opts.DefaultAdmins, authnUser.Email)`, a byte comparison. That is the bootstrap path of an empty instance: an admin who writes `Jean.Dupont@corp.tld` in `XOLO_HTTP_AUTHN_DEFAULT_ADMINS` while the provider returns `jean.dupont@corp.tld` gets a 403 and no way out. The provisioning service already lowercases those addresses, so the two paths disagree. `slices.ContainsFunc` with `strings.EqualFold`, or normalizing `opts.DefaultAdmins` when the middleware is built, fixes it.

## The decline link on /join is unreachable for the invitees this PR creates

`join/handler.go:75` builds a decline form pointing at `POST /no-org/invitations/{tokenID}/decline`, mounted behind `isActive` (`webui/handler.go:94`), while `/join/` itself is not. With `ActiveByDefault` at its `false` default, exactly the population the new exemption provisions, the invitee can accept but the decline button sends them to the inactive-account page. Mount the decline route outside `isActive`, or hide the form for an inactive user.

## Normalize the address on write, not only on read

`internal/http/handler/webui/org/invites.go:139-141` stores `r.FormValue("invitee_email")` as is, with no `TrimSpace`. An address pasted from a mail client with a trailing space still escapes `LOWER(invitee_email) = LOWER(?)` and the invitation stays invisible to its recipient, which is the silent symptom this PR sets out to remove.

Writing `strings.ToLower(strings.TrimSpace(email))` also settles the index question: `InviteeEmail` carries `gorm:"index"` (`internal/adapter/gorm/invite.go:14`), and wrapping the column in `LOWER()` makes the predicate non-sargable, so SQLite and PostgreSQL both scan `invite_tokens`. The table is short today, but this query now sits on the authentication path and runs again on every refused attempt, held back only by the rate limiter. Was read-only normalization a deliberate choice to avoid migrating existing rows?

## Smaller points

- `isInvited := hasPendingInvite(...)` at `middleware.go:91` is assigned before the guard on line 96, so Go's short-circuit does not apply and the lookup runs even with `AutoCreateUsers` at its `true` default. It is confined to the unknown-identity branch, so the cost is bounded, but a closure evaluated only when `!opts.AutoCreateUsers && !isDefaultAdmin && !isApplication` costs nothing.
- The bridge covers revocation, foreign tenant and open invitation, but not the case the comment on `middleware.go:206-209` invokes to justify calling `IsInviteValid`: a targeted invitation with `MaxUses=1` and `UsesCount=1`. A subtest for that, and one where the store returns an error, would pin both.
- The exemption trusts the provider's `email` claim, like `DefaultAdmins` above it. Consistent with what exists, and worth one sentence in `docs/fr/administration/organisation/invitation/invitation.md`: on an instance exposing several providers, with `ActiveByDefault=true`, anyone who gets an identity carrying the invited address joins the organization without ever receiving the link.

## Out of scope, worth its own issue

`declineInvitation` (`internal/http/handler/webui/no_org.go:70`) fetches the invitation by ID and deletes it when it is targeted, without comparing `invite.InviteeEmail()` to `user.Email()`, a check `acceptInvite` does perform. Any active authenticated user who knows an invitation ID destroys someone else's invitation, irreversibly and silently. The IDs are `xid`, so not guessable, but they travel in the clear inside `/join/{tokenID}` links.

PR #54, opened 2026-09-16
fix(config): trim XOLO_HTTP_BASE_URL once at config parse

### LINKED ISSUE #28: config: XOLO_HTTP_BASE_URL is validated trimmed but used untrimmed
Raised while reviewing #26, not blocking it.

`validateMultitenantBaseURL` (`internal/config/config.go`) parses a trimmed copy of the value:

```go
parsed, err := url.Parse(strings.TrimSpace(baseURL))
```

but `c.HTTP.BaseURL` itself is never normalised, and that untrimmed value is what reaches `http.WithBaseURL`, `newTenantBaseURLResolver` and `oidcCallbackURL`.

A stray space or newline is easy to pick up from a `.env` file or a ConfigMap. In multi-tenant mode it clears config validation and then fails further along in `newTenantBaseURLResolver` ("could not parse base url"), at a point that says nothing about where the value came from. In single-tenant mode, where the check does not run at all, it propagates quietly into every generated URL, OAuth redirect_uri included, and the login then fails at the IdP over a mismatch nobody can see.

## What to do

Trim `c.HTTP.BaseURL` once, when the config is parsed, so the value that gets validated is exactly the value everything else uses.

Padiwa:
Closes #28 
### Change

Config.Validate now runs XOLO_HTTP_BASE_URL through a small normalizeBaseURL helper once, before any validation runs. The helper trims surrounding whitespace and a trailing slash.

The value that gets validated is therefore exactly the value that every downstream consumer (http.WithBaseURL, newTenantBaseURLResolver, oidcCallbackURL) sees. The previous defensive strings.TrimSpace inside validateMultitenantBaseURL is removed as redundant, with a comment that documents the new contract.

### Beyond the issue's stated scope

normalizeBaseURL also strips a trailing /. This is not what issue #28 asked for, but the change is benign and arguably beneficial:

- oidcCallbackURL already had to defensively TrimRight the base URL for the same reason.
- url.JoinPath calls downstream would otherwise produce a stray // in built URLs.
- The return "/" fallback is what makes this safe for the envDefault "/" and any whitespace-only input (" ", a lone newline, a stray ConfigMap line). Without it, TrimRight("/", "/") would collapse those shapes to "" and url.Parse("").JoinPath(...) would drop the leading slash, breaking every redirect built from the base (see no_org.go, join/handler.go, authn/oidc/provider.go, org/invites.go).

Tests

- TestNormalizeBaseURL — pins the trimming behaviour across twelve cases: the envDefault "/" and a whitespace-only value fall back to "/", an absolute URL stays absolute (with or without trailing slash), leading/trailing/surrounding whitespace is trimmed, and a literal "" (only reachable by constructing config.Config directly or via empty `${VAR}` expansion through the `expand` tag) stays "". Each case asserts both the stored value AND url.Parse(stored).JoinPath("/x").String() — the latter is the contract downstream consumers rely on, and the assertion that would have caught the whitespace-only regression in this PR's first review.
- TestValidateMultitenantBaseURL — two cases exercise the issue #28 symptom at the multi-tenant validation gate: a padded absolute URL must be accepted, and a padded schemeless URL must still be rejected.
- TestNormalizeBaseURLFixesMultitenantValidationBypass — runs with `Multitenancy{Enabled: true, HostPattern: "...", DefaultTenantSlug: "default"}` and verifies that a raw input like `" https://xolo.example.com/ "` reaches Validate as the canonical `https://xolo.example.com`, parses with non-empty Scheme and Host, and that JoinPath("/x") yields an absolute URL the browser can resolve against the root. Without normalizeBaseURL this test reproduces the original symptoms from issue #28.

The two contract assertions (Scheme/Host non-empty AND JoinPath("/x") prefix) intentionally go beyond asserting stored-string equality because net/url's handling of relative bases (in particular url.Parse("")) has caused exactly this class of failure in past Go versions.

Operator-visible change

None in the success path for any operator who sets XOLO_HTTP_BASE_URL to a non-empty value, including a single blank character. In the failure path:

- Multi-tenant operators who had a stray space in their XOLO_HTTP_BASE_URL no longer get the cryptic "could not parse base url" or "must be absolute" error at first request time — the value is normalised at startup and the validation either passes or fails with a clearer signal. Locked in by the new padded-absolute and padded-schemeless cases in TestValidateMultitenantBaseURL.
- Single-tenant operators no longer get silent mismatches at the IdP on the OAuth redirect_uri — the value flows canonically.
- Operators who relied on a trailing / being preserved now get the slashless form (e.g. `https://xolo.example.com/` -> `https://xolo.example.com`). All current consumers treat both forms identically, but anyone with a custom consumer that distinguished them should review.

## Documented limitation (unreachable from env loading)

A literal "" stored on HTTP.BaseURL still trips url.Parse("").JoinPath(...) = "x" with no leading slash. This is only reachable by constructing config.Config programmatically (the tests do) or by feeding an empty `${VAR}` expansion through the `expand` directive on the env tag. caarlos0/env v11.3.1 substitutes envDefault:"/" for the explicit empty case before normalizeBaseURL sees it, so a `XOLO_HTTP_BASE_URL=""` deployment yields "/" today; the limitation is pinned by the "empty stays empty" test case as future-proofing against alternate env loaders or shell-level expansion that would otherwise reintroduce it.


Bornholm (changes requested):
The fix itself is right. Normalizing once at the top of `Validate`, before `validateMultitenantBaseURL`, means the value that gets validated is the one `http.WithBaseURL`, `newTenantBaseURLResolver` and `oidcCallbackURL` receive. That is what #28 asks for. The trailing slash trim is where it goes wrong.

`strings.TrimRight(..., "/")` turns the `envDefault:"/"` of `internal/config/http.go:6` into the empty string, so it lands on every single-tenant deployment that never sets the variable. The comment claims the consumers treat an empty base as equivalent to `/`. They do not:

```go
url.Parse("").JoinPath("/no-org").String()  // "no-org"
url.Parse("/").JoinPath("/no-org").String() // "/no-org"
```

`url.Parse("")` leaves `Path` empty, so `joinPath` takes its relative branch and returns a path with no leading slash. Absolute bases are fine, `url.String()` puts the slash back as soon as `Host` is set.

What that breaks under the default config, with every URL resolved by the browser against the current request directory instead of the root:

- `internal/http/handler/webui/no_org.go:70` and `:89`, the redirects after accepting or declining an invitation
- `internal/http/handler/webui/join/handler.go:71-72`, `LoginURL` and `DeclineURL`
- `internal/http/middleware/authn/oidc/provider.go:129`, which redirects to `baseURL.String()`, now empty
- `internal/http/handler/webui/org/invites.go:168`, the invite link people paste into a chat

For a bugfix PR I would keep the `TrimSpace` #28 asks for and leave trailing slashes to the consumers, since `oidcCallbackURL` already strips one. If you want to keep the trim, never return an empty string from a non-empty base:

```go
func normalizeBaseURL(raw string) string {
	trimmed := strings.TrimSpace(raw)
	if trimmed == "" {
		return ""
	}
	if cleaned := strings.TrimRight(trimmed, "/"); cleaned != "" {
		return cleaned
	}
	return "/"
}
```

Two points on the tests:

- They assert the stored string and nothing else, and `internal/config/base_url_test.go:92-95` pins `"/" -> ""` as the expected result. An assertion on `url.Parse(conf.HTTP.BaseURL).JoinPath("/x").String()` would have caught this.
- `TestNormalizeBaseURL` and `TestNormalizeBaseURLFixesSingleTenantBypass` both build `config.Multitenancy{DefaultTenantSlug: "default"}` without `Enabled: true`, so `validateMultitenantBaseURL` returns immediately. The multi-tenant case from #28, a padded value that passed validation and then failed in `newTenantBaseURLResolver`, is never exercised. The `DefaultTenantSlug` field suggests otherwise.

Minor: the comment on `internal/config/config.go:157` closes with a curly `”` instead of `""`, which makes the sentence unreadable. It disappears if the trim is dropped.

Bornholm (changes requested):
Thanks for the follow-up — the blocking points from the previous round are addressed: `normalizeBaseURL` now runs once at the top of `Validate()`, so `http.WithBaseURL`, `newTenantBaseURLResolver` and `oidcCallbackURL` all receive the exact validated value; the `return "/"` fallback keeps the `envDefault:"/"` from collapsing to `""`; the tests assert the `url.Parse(stored).JoinPath("/x").String()` contract instead of just the stored string; and the multi-tenant case is now exercised with `Multitenancy.Enabled = true`.

One correctness gap left, plus two comments that say the opposite of what the code does.

### 1. A whitespace-only value normalizes to `""` instead of falling back to `"/"` (blocking)

`internal/config/config.go:161-170` — `normalizeBaseURL` returns `""` as soon as `strings.TrimSpace(raw)` is empty, so an input made only of blanks (`" "`, a newline, a ConfigMap or `.env` line containing a single space) lands on the one shape the code cannot handle.

This is reachable from the environment: `caarlos0/env` v11.3.1 only substitutes `envDefault` when the variable is defined *and strictly empty* (`env.go:640-645`, `case exists && value == "" && defExists`). `" "` is not empty for `getOr`, so it goes straight through the parsing. In single-tenant mode `validateMultitenantBaseURL` returns immediately, the empty value propagates silently, and `url.Parse("").JoinPath("/no-org").String()` yields `no-org` with no leading slash — the browser then resolves redirects and invite links against the current request's directory instead of the root. That's exactly the silent-failure class this PR removes for every other shape. In multi-tenant mode the absolute-URL check catches it at startup.

Suggested fix — distinguish the raw empty from the post-trim empty so the blank input falls into the existing fallback:

```go
func normalizeBaseURL(raw string) string {
	if raw == "" {
		return ""
	}
	if cleaned := strings.TrimRight(strings.TrimSpace(raw), "/"); cleaned != "" {
		return cleaned
	}
	return "/"
}
```

and add a `" "` case to the table expecting `want "/"` / `wantURL "/x"`. No test currently covers this input.

### 2. `TestNormalizeBaseURL`'s doc comment contradicts what the test pins

`internal/config/base_url_test.go:108-113` — the comment ends with "The consumers all treat an empty stored value as equivalent to `/` exactly so that `http.WithBaseURL("").String()` and friends keep producing `/x`." The table pins `"" -> wantURL "x"`, i.e. the exact opposite, and the trailing comment in the same function ("the consumers then hand `x` (no leading slash) to the browser") says so too. A maintainer reading only the first comment would conclude an empty base is safe. Please rewrite it: consumers do *not* treat an empty stored value as `/`; the case is pinned to `x` to document the limitation, and it is the `"/"` fallback that protects the single-tenant default.

### 3. The "known limitation" is overstated

`internal/config/base_url_test.go:117-130` and the PR description warn operators who explicitly set `XOLO_HTTP_BASE_URL=""`. With the pinned dependency that value never reaches `normalizeBaseURL`: `env.go:645` substitutes the `envDefault`, and `internal/config/http.go:6` carries `envDefault:"/"`, so an explicitly empty variable yields `/`. The empty input is only reachable by building `config.Config` directly (what the tests do) or through an empty `${VAR}` expansion, since the tag carries `expand`. The unit pin is still legitimate — just reword it and the "Known limitation" paragraph accordingly.

Related, in the PR description: "Operators who relied on a trailing `/` being preserved get an empty string instead" is no longer true since the fallback was added — no non-empty input collapses to `""`; trimming the trailing slash gives the slashless form (`https://h/` -> `https://h`).

Point 1 is the only blocking one; 2 and 3 are documentation fixes.

Bornholm (changes requested):
New round on `a15f811`. The two blocking points from last time are gone: `normalizeBaseURL` runs once at the top of `Validate()`, whitespace-only inputs now fall into the `"/"` fallback with three table cases pinning `wantURL "/x"`, and the test comments no longer contradict the table. Good.

One correctness gap left, and it is the point I got wrong last round.

## The literal-empty passthrough is reachable from the environment

`internal/config/config.go:162-176` keeps `if raw == "" { return "" }`, justified in the comment as useful for code building a `config.Config` directly. I said in my previous review that an empty value could only come from direct construction. That was wrong, and the dependency source says so: `HTTP.BaseURL` carries `env:"BASE_URL,expand"` (`internal/config/http.go:6`), and caarlos0/env v11.3.1 runs `os.Expand` *after* `getOr` has substituted the `envDefault` (`env.go:585-597`).

So `XOLO_HTTP_BASE_URL='${PUBLIC_URL}'` with `PUBLIC_URL` unset, which is ordinary ConfigMap and compose templating, expands to `""`, reaches `normalizeBaseURL`, survives it, and in single-tenant mode skips `validateMultitenantBaseURL` entirely. From there `url.Parse("").JoinPath("/x").String()` gives `x` with no leading slash, and the browser resolves it against the current request directory. That hits `no_org.go` redirects, `LoginURL` and `DeclineURL` in `join/handler.go`, the OIDC callback redirect, and the invite link in `org/invites.go`. It is the same silent failure this PR removes for every other shape.

No production code builds `config.Config` directly, so nothing depends on `""` surviving. Dropping the special case is enough:

```go
func normalizeBaseURL(raw string) string {
	if cleaned := strings.TrimRight(strings.TrimSpace(raw), "/"); cleaned != "" {
		return cleaned
	}
	return "/"
}
```

Then flip the "empty stays empty" case to `want "/"` / `wantURL "/x"`, and the "documented limitation" paragraph goes away instead of being pinned as intended behaviour.

## The padded cases in TestValidateMultitenantBaseURL do not test the fix

`internal/config/base_url_test.go:51-74`. Both cases assert only whether `Validate()` returns an error, never the stored value. At the merge base, `validateMultitenantBaseURL` already parsed `url.Parse(strings.TrimSpace(baseURL))` (`8c5265e`, `internal/config/config.go:144`), so the padded absolute URL already passed and the padded schemeless one already failed with the same message. Both cases behave the same before and after the fix, yet their comments call them an issue #28 regression test. Only `TestNormalizeBaseURLFixesMultitenantValidationBypass`, which asserts the stored value, catches the regression. Someone trusting the comments could delete that test.

Either assert the stored value in the padded-absolute case, or reword the comments to say they pin the multi-tenant accept and reject contract after normalization.

## TestNormalizeBaseURLFixesMultitenantValidationBypass's comment says the opposite of what url.Parse does

`internal/config/base_url_test.go:226-233`. The comment claims `url.Parse` on a padded absolute URL returns a URL with `Scheme` and `Host` populated, so the lenient call accepts the input. It does not. `url.Parse(" https://xolo.example.com/ ")` returns `first path segment in URL cannot contain colon` and populates neither field. The real pre-fix symptom for that input was the "could not parse base url" wrap in `newTenantBaseURLResolver`, which the comment on `Validate` (`internal/config/config.go:118-121`) gets right.

Reword it: the padded absolute URL passed `validateMultitenantBaseURL`, which parsed a trimmed copy, then failed later in `newTenantBaseURLResolver`. The "must be absolute" branch is what a value parsing with empty scheme and host, such as `xolo.example.com`, hits.

## Minor

`internal/config/config.go:159-161`: the comment writes `TrimRight("/", "/")` where the code passes `TrimRight(strings.TrimSpace(raw), "/")`. True for the `"/"` input, false for the whitespace shapes the same sentence lists.

## Out of scope, worth a follow-up issue

The trailing-slash trim is not neutral for one consumer. The GitHub provider registers `Issuer: conf.HTTP.BaseURL` (`internal/setup/authn_oidc_handler.go:106`) whenever the base URL is neither `""` nor `"/"`, and that string reaches `jwt.WithIssuer` (`oidctoken/authenticator.go:129`), an exact comparison against the token `iss` claim. Unlike the path-joining consumers, `https://xolo.example.com` and `https://xolo.example.com/` are two different issuers there. An operator who set the trailing-slash form to match their tokens would see validation start failing after the upgrade, with an error pointing at the token rather than the config. It is an unusual setup and I could not confirm a deployment relying on it, but the PR description's claim that "all current consumers treat both forms identically" does not hold.

Also worth noting: `oidcCallbackURL` still runs its own `strings.TrimRight(baseURL, "/")` (`authn_oidc_handler.go:44`), which is now redundant and works against the single-source-of-truth contract this PR argues for in `config.go:186-192`. And the French configuration reference still describes `XOLO_HTTP_BASE_URL` as a plain public URL, with no mention of the normalization operators now get at startup.

Only the first point blocks.

PR #59, opened 2026-09-17
fix(tenant): reject empty host and use configured base URL in Canonic…

### LINKED ISSUE #30: tenant: CanonicalHost returns an empty or forged host unvalidated in single-tenant mode
Raised while reviewing #26. Not reachable today, so no user-visible bug: filed so the code matches what it promises.

`Resolver.CanonicalHost` (`internal/http/middleware/tenant/middleware.go`) validates nothing on the single-tenant branch:

```go
if !r.multiTenant {
	return strings.ToLower(stripPort(host)), true
}
```

So `CanonicalHost("")` answers `("", true)`, and any forged Host comes straight back out, lowercased. The doc comment right above the method states the opposite:

> Callers that build public URLs from a request host use it so that neither a forged host, its casing nor a client-supplied port can leak into a generated URL.

The only caller today is `newTenantBaseURLResolver`, which `NewHTTPServerFromConfig` installs only when `conf.Multitenancy.Enabled`, so the branch is dead. A future single-tenant caller would build a URL with an empty authority (`perHost.Host = ""`) or copy a client-supplied Host into a redirect_uri.

## What to do

Return `false` when the normalised host is empty. On the single-tenant branch, prefer the host of the configured base URL over the request's, which is what a caller asking for a canonical host actually wants. Either way the contract in the comment holds on both branches.

Padiwa:
## What                                                                                                                                                                                                          
                                                                                                                                                                                                                    
   Fixes a host-header injection vulnerability in the tenant middleware's `CanonicalHost` function when running in single-tenant mode.                                                                              
                                                                                                                                                                                                                    
   ### Problem                                                                                                                                                                                                      
                                                                                                                                                                                                                    
   In single-tenant mode, `Resolver.CanonicalHost(host)` previously echoed the client-supplied `Host` header (lowercased, port stripped). Any URL builders that rely on `CanonicalHost` to produce a safe authority 
 would therefore leak whatever host an attacker chose to put in their request. A request with `Host: evil.example.com` could end up influencing the host used in links and redirects returned by the server.        
                                                                                                                                                                                                                    
   ### Fix                                                                                                                                                                                                          
                                                                                                                                                                                                                    
   The single-tenant canonical host now comes from the configured public base URL (`XOLO_HTTP_BASE_URL`), not from the request. The resolver records this value once at construction time:                          
                                                                                                                                                                                                                    
   - `Resolver` gains a `singleTenantHost` field, populated by `NewResolver` from the configured base URL.                                                                                                          
   - A new helper, `canonicalHostFromBaseURL`, parses the base URL with `net/url`, lower-cases the host and strips the port. An empty, relative or malformed base URL yields `""`.                                  
   - `CanonicalHost` is rewritten so that:                                                                                                                                                                          
     - In **multi-tenant** mode, behavior is unchanged: the slug is extracted via `slugFromHost` and the canonical host is rebuilt from the configured `hostPrefix + slug + hostSuffix` pattern.                    
     - In **single-tenant** mode, the function now returns `singleTenantHost` regardless of the request's `Host` header. When no base URL is configured, it returns `("", false)` so callers cannot forge a URL     
 from an empty authority.                                                                                                                                                                                           
                                                                                                                                                                                                                    
   The documented contract is: on both branches, neither a forged host, its casing, nor a client-supplied port can leak into a generated URL.                                                                       
                                                                                                                                                                                                                    
   ### API change                                                                                                                                                                                                   
                                                                                                                                                                                                                    
   `tenant.NewResolver` now takes a third argument:                                                                                                                                                                 
                                                                                                                                                                                                                    
   ```go                                                                                                                                                                                                            
   func NewResolver(store port.TenantStore, conf config.Multitenancy, baseURL string) *Resolver                                                                                                                     
   ```                                                                                                                                                                                                              
                                                                                                                                                                                                                    
   All call sites are updated in this PR.                                                                                                                                                                           
                                                                                                                                                                                                                    
   ## Why                                                                                                                                                                                                           
                                                                                                                                                                                                                    
   Closes #30.                                                                                                                                                                                                      
                                                                                                                                                                                                                    
   The resolver was meant to be the single source of truth for "what host may appear in a URL we generate", but the single-tenant branch was simply echoing the request. Anchoring it to the operator-controlled    
 base URL removes the ambiguity and the attack vector in one move.                                                                                                                                                  
                                                                                                                                                                                                                    
   ## Changes                                                                                                                                                                                                       
                                                                                                                                                                                                                    
   - `internal/http/middleware/tenant/middleware.go` (+55/−16): new `singleTenantHost` field, new `canonicalHostFromBaseURL` helper, rewritten `CanonicalHost`, updated `NewResolver` signature.                    
   - `internal/http/middleware/tenant/middleware_test.go` (+94/−14): update existing tests to the new signature, adjust the previous "single tenant matches anything" case to expect rejection, replace the         
 redundant "single tenant ignores the request host entirely" case with one that exercises lowercase + port-strip on the configured base URL, and add `TestResolverCanonicalHostSingleTenant` covering the new       
 contract (configured host returned instead of the request host, lowercase + port-strip, empty request host, relative base URL rejected, missing base URL rejected).                                                
   - `internal/setup/http_server.go` (+1/−1): pass `conf.HTTP.BaseURL` to `tenant.NewResolver`.                                                                                                                     
   - `internal/setup/base_url_resolver_test.go` (+1/−1): update call to the new `NewResolver` signature.


Bornholm (changes requested):
Reviewed at `4d6d273`, still in draft. I ran a multi-agent pass and then checked the findings myself, including running the parsing code against the cases below.

The fix is the right one. Anchoring the single-tenant canonical host to `XOLO_HTTP_BASE_URL` and returning `("", false)` when none is configured makes the method match the contract written above it, on both branches. Moving the derivation into `NewResolver` so it happens once rather than per request is the right call too.

Two things before this leaves draft. The first is a real bug, the second is about how the change is described.

## canonicalHostFromBaseURL drops the brackets on a literal IPv6 host

`stripPort` goes through `net.SplitHostPort`, which unbrackets. So the output changes depending on whether the base URL carries a port:

```
http://[::1]:8080  ->  parsed.Host = "[::1]:8080"  ->  canonical = "::1"
http://[::1]       ->  parsed.Host = "[::1]"       ->  canonical = "[::1]"
```

A caller that assigns `"::1"` to `url.URL.Host` builds `http://::1/...`, which is not a valid URL. `parsed.Hostname()` unbrackets the same way, so that is not the fix either. Re-bracket the host when it contains a `:`, or use `net.JoinHostPort` and keep the port.

Rare for a public base URL, and no caller reaches this branch today, so I would not hold the PR on it if it were the only thing. But it is cheap to fix while the code is fresh, and the inconsistency between the two forms is the kind of thing that costs an afternoon later.

## The description claims an exploitable vulnerability, #30 says otherwise

The description opens with "Fixes a host-header injection vulnerability" and talks about an attacker choosing the host that ends up in links and redirects. #30, which this PR closes, says the opposite in its first line:

> Not reachable today, so no user-visible bug: filed so the code matches what it promises.

I checked, and #30 is right. `CanonicalHost` has exactly one caller in the tree, `newTenantBaseURLResolver` at `internal/setup/http_server.go:343`, and `NewHTTPServerFromConfig` only installs it under `if conf.Multitenancy.Enabled`. The single-tenant branch this PR rewrites is not reachable in a running server. Nothing was leaking.

Please reword the description to match. This is hygiene that closes the branch before someone wires up a caller, which is worth doing, and the PR is good work. But a PR that says "fixes a host-header injection" gets read as a live fix. It lands in release notes that way, it changes how urgently someone backports it, and if it ever reaches a security advisory it will not survive the first person who checks the call graph.

## Not asking for a change

The test additions cover the branch properly, including the empty and relative base URL cases. The `NewResolver` signature change with every call site updated is fine.

One question, no change requested. In single-tenant mode a missing or relative base URL is accepted silently at construction and only shows up as `ok=false` at the first `CanonicalHost` call. Multi-tenant mode validates the base URL at startup in `validateMultitenantBaseURL`. Once something does call this branch, would the same startup check be worth adding here, so an operator who forgot `XOLO_HTTP_BASE_URL` finds out at boot instead of at runtime?

Bornholm (changes requested):
Reviewed at `8a173b9`. The IPv6 fix lands, CI is green on all five jobs, and the code side of this PR is in good shape. What is blocking is the description, which is the same ask as the previous round.

## What is now closed

The IPv6 bracket bug is fixed. I re-ran both forms rather than reading for it: `http://[::1]:8080` and `http://[::1]` now both yield `[::1]`, so the same operator-visible setting no longer produces two different shapes depending on whether a port is present. The two regression cases added to `TestResolverCanonicalHostSingleTenant` cover both inputs.

## The description still claims an exploitable vulnerability

It opens with "Fixes a host-header injection vulnerability" and describes an attacker whose host ends up in generated links and redirects. Issue #30 opens with the opposite: "Not reachable today, so no user-visible bug."

I checked the call graph again at this revision. `CanonicalHost` has exactly one caller outside tests, `newTenantBaseURLResolver` at `internal/setup/http_server.go:343`, and `WithBaseURLResolver` is installed only under `if conf.Multitenancy.Enabled` (`http_server.go:342-349`). The single-tenant branch this PR rewrites is not reachable from any request on a running server. No `Host` header can reach a generated URL through it, forged or not.

The code change is still worth making. Anchoring the branch to the operator-controlled base URL before a caller is wired to it is the right order to do things in. But that is hardening, not a fix for an active flaw, and the current wording is what goes into the merge message, the release notes and any backport decision. If it ever reached a security advisory it would not survive the first call-graph check.

Rewrite the opening as closing the branch before a caller uses it, and drop "removes the attack vector in one move" from the Why section. Keep the link to #30, which already states the reachability correctly.

## Collapse the duplicated bracketing logic

`canonicalHostFromBaseURL` at `middleware.go:218-225` does the bracketing twice. The `if strings.HasPrefix(parsed.Host, "[") && !strings.HasPrefix(host, "[")` branch re-adds the brackets inline, and its own `else` calls `bracketIfIPv6`, which does the same thing. I ran both paths against `http://[::1]:8080`, `http://[::1]`, `//xolo.example.com`, `http://user:pass@host.example.com` and `https://xolo.example.com:443`. Identical output on all five.

The whole block is one line:

```go
host := bracketIfIPv6(stripPort(parsed.Host))
```

Worth doing now because `bracketIfIPv6` has no other caller. A helper introduced for future callers, then bypassed by the only current one, is the kind of thing that gets deleted as dead six months later.

## Rename the port test case

`middleware_test.go:285-293` is named "the request port is ignored" and passes `host: "xolo.example.com:9999"`. The single-tenant branch no longer reads its `host` argument at all, so the request port cannot reach the output by construction and the case cannot regress on a missing port strip. It still pins the contract, so keep it, but the name will mislead the next reader into thinking port stripping is covered here. The neighbouring case "lowercases and strips the port from the configured base url" is the one that tests that.

"the request port is irrelevant when a base url is configured" matches both the behaviour and the case you already named "the request host is irrelevant".

## Not asking for a change

A missing or relative `XOLO_HTTP_BASE_URL` is accepted silently at startup in single-tenant mode and only refused at the first `CanonicalHost` call, where multi-tenant validates up front through `validateMultitenantBaseURL`. The asymmetry has no consequence while the branch has no caller. It becomes a silent startup failure the day someone wires one, so it belongs in whatever issue does that wiring rather than here.

A protocol-relative base URL like `//xolo.example.com` parses with `Host` set and would be accepted as the canonical host, with no scheme check. Same reasoning: not reachable, worth a line in the follow-up issue.

Bornholm (approved):
## PR #59 — Approved

The single-tenant branch of `CanonicalHost` now derives the canonical host from `XOLO_HTTP_BASE_URL` and returns `("", false)` when no absolute base URL is configured. The IPv6 bug is fixed, the duplicate bracketing logic is collapsed to a single `bracketIfIPv6` call, and the test coverage is complete.

The two open points were already explicitly deferred by the human reviewer to a follow-up issue:
- `BaseURL` defaults to "/" in single-tenant mode — the error surfaces only at the first call to `CanonicalHost`, not at startup. A startup validation matching the multi-tenant pattern belongs in whatever issue wires up a single-tenant caller.
- A protocol-relative URL (`//xolo.example.com`) is accepted without a scheme check. Not reachable in production today, but worth a line in the follow-up issue.

PR #68, opened 2026-09-18
fix(webui): allow renaming virtual models (personal and org)

### LINKED ISSUE #60: Custom model can't be renamed
### What happened?

When trying to rename a custon model, it seems that you can't input any value in the model name field

### Steps to reproduce

1. chose a custom model
2. click Rename button
3. try to rename th model

### Xolo version

0.71

### Environment

Chromium / Firefox

### Relevant logs

```text
on access to /profile/personal-models/.../edit

VM58:2 Uncaught TypeError: Cannot read properties of undefined (reading 'startTime')
    at X.reportAllChanges (<anonymous>:2:19627)
    at <anonymous>:2:14832
    at <anonymous>:2:313
    at d (<anonymous>:2:6931)
    at <anonymous>:2:7118
    at <anonymous>:2:2912
    at n.timeout (<anonymous>:2:6432)
X.reportAllChanges @ VM58:2
(anonymous) @ VM58:2
(anonymous) @ VM58:2
d @ VM58:2
(anonymous) @ VM58:2
(anonymous) @ VM58:2
n.timeout @ VM58:2
requestIdleCallback
x @ VM58:2
f @ VM58:2
(anonymous) @ VM58:2
```

Padiwa:
## What

Closes #60.

The "Rename" button on a virtual model (personal or organisation) led to a form where the `name` field was `Readonly: true`, with no editable value. The `POST` that followed ignored the name entirely. On the organisation side the button was labelled "Rename" but renamed nothing — exact same symptom.

Three coordinated changes expose the rename:

- The domain model exposes `SetName(string)` on `BasePersonalVirtualModel` and `BaseVirtualModel`.
- The GORM wrappers implement it too, so `SavePersonalVirtualModel` and `SaveVirtualModel` already propagate the new name. The existing unique indexes (`idx_pvm_user_name`, `idx_org_name`) maintain the invariant at the database level; they remain the safety net for any race between the uniqueness pre-check and the save.
- The web handlers `updatePersonalModel` and `updateVirtualModel` read `name`, check that it is not empty, check uniqueness through `Get...ByName` while ignoring the current row, then call a clean `mutable{ SetName, SetDescription, SetUpdatedAt }` interface. The validation rule is taken as-is from `create`: non-empty and unique per user (or per org). No extra regex.

The edit pages (`personal_models_page.templ`, `virtual_models_page.templ`) now render the field editable, with a label and helper text that describe the rename and the uniqueness constraint. On collision, empty name, or save failure the handler redirects back to the same page with `?error=exists`, `?error=name_required`, or `?error=update_failed`. `getEdit...Page` propagates the query string to `...FormVModel.Error` so `common.FlashError` renders the right message. The `name_required` flash code is mapped in `common/flash.templ` to "Le nom est obligatoire." and is consistent across the four create/update handlers.

## Latent bug

The historical `updatePersonalModel` did `vm.(*model.BasePersonalVirtualModel)` while `GetPersonalVirtualModelByID` returns `*wrappedPersonalVirtualModel`. The assertion would always have failed in production (HTTP 500), but the "Save" button on the edit form is not the path the users in the issue took — they could not type a name at all — so the bug went unnoticed. The new handler uses a `mutable` interface consistent with the JSON API `handleUpdatePersonalVM`.

## Cross-org access

An admin of org A who knew or guessed the XID of a model belonging to org B could previously open its edit page, modify it, delete it, or open its pipeline editor (the editor exposes the full graph — node names, plugin configuration). RBAC on the routes checks `PermVirtualModelsRead` / `PermVirtualModelsWrite` for the request's org, not for the model's org, and the store loads by ID only — the ownership guard was missing. A `vm.OrgID() != org.ID() { http.NotFound; return }` check is added to `getEditVirtualModelPage`, `updateVirtualModel`, `deleteVirtualModel` (which now also resolves the org from the slug), and `getPipelineEditorPage`. The personal handlers (`getEditPersonalModelPage`, `updatePersonalModel`, `deletePersonalModel`, `getPersonalPipelineEditorPage`) already enforced the equivalent `UserID()` check.

The same vulnerability class affects five middleware handlers (`getEditMiddlewarePage`, `updateMiddleware`, `toggleMiddleware`, `deleteMiddleware`, `getMiddlewarePipelineEditorPage`). Out of scope here; tracked in #70.

## Out of scope

- The JSON API (`PUT /api/personal-models/{vmID}`, `PUT /api/orgs/{slug}/virtual-models/{id}`) still does not support renaming. The local `mutable` interfaces in the JSON handlers do not include `SetName`, and the `Name` field in the request body is silently ignored. To be extended if usage justifies it.
- The `Cannot read properties of undefined (reading 'startTime')` JS error reported in the ticket comes from the React scheduler and concerns the pipeline editor (`/.../pipeline`), not the edit page. Not reproduced here.
- No schema migration: the `name` column already existed and stays NOT NULL + indexed. Renaming is a standard `UPDATE` on the existing row.
- No handler-level tests for the new ownership checks. Intentionally deferred — natural to land alongside the middleware fix in #70.
- No helper to dedup the `vm.OrgID() != org.ID()` check (currently written four times in `org/virtual_models.go`). Deferred so a single helper can cover both this PR's surface and #70's.

## Notes for review

- The choice to validate uniqueness in the handler rather than at the store layer is consistent with `create` and lets the handler phrase the right error message; the unique index remains the safety net and preserves correctness even under a concurrent-rename race.
- No schema change, no public API change, no new dependency. The PR is purely additive on the model side (one new setter) and replaces the historical `if err == nil && existing != nil` pattern with a `switch` on `errors.Is(err, port.ErrNotFound)` to distinguish a real collision from a store error.
- `make generate`, `make build`, `go test ./...`, and `make test` (including `pkg/pluginsdk`) pass with no FAIL. `go vet ./internal/http/handler/webui/...` is silent.


Bornholm (changes requested):
Reviewed at `1085f53`. I ran a multi-agent pass and then checked every finding against the tree myself, so what follows is what I could confirm by reading the code at head.

The rename works and the write path is sound. `updatePersonalModel` asserting `*model.BasePersonalVirtualModel` against a `*wrappedPersonalVirtualModel` was a real 500 on every save, and the `mutable` interface is the right fix. The four `vm.OrgID() != org.ID()` guards are in place in `getEditVirtualModelPage`, `updateVirtualModel`, `deleteVirtualModel` and `getPipelineEditorPage`, `deleteVirtualModel` now resolves the org from the slug before the guard, and the personal side keeps its four `UserID()` checks. Replacing `if err == nil && existing != nil` with a switch on `errors.Is(err, port.ErrNotFound)` is a genuine improvement over what was there.

Two things to fix before merge. Neither is a correctness bug, and the first one is the one that motivates the request.

## Save does not map a duplicate key the way Create does

`CreateVirtualModel` and `CreatePersonalVirtualModel` both catch `gorm.ErrDuplicatedKey` and return `port.ErrAlreadyExists` (`virtual_model_store.go:16`, `personal_virtual_model_store.go:16`). `SaveVirtualModel` and `SavePersonalVirtualModel` do not. They go through `clause.OnConflict{Columns: id, UpdateAll: true}` and return the raw GORM error.

The unique index is `idx_org_name`, not the id, so a rename that collides does not match the `ON CONFLICT` target and the driver error comes straight back up. The handler has no sentinel to test, falls into its default branch and redirects to `?error=update_failed`. The user gets "L'enregistrement a échoué" when the real answer is "that name is taken".

This only fires when two renames race, since the handler pre-check catches everything else. Your note says the unique index stays the safety net under a concurrent rename, and that is true for the data. It is the message the user reads that is wrong. And this path was unreachable before, because renaming was not possible, so the PR is what makes it reachable.

Map `gorm.ErrDuplicatedKey` to `port.ErrAlreadyExists` in both `Save*`, then send that sentinel to `?error=exists` in the two update handlers.

## The `existing.ID() != vm.ID()` branch cannot run

In `updateVirtualModel` and in `updatePersonalModel`, the block is guarded by `if name != vm.Name()`. `GetVirtualModelByName(org.ID(), name)` matches on `=`, and both SQLite and PostgreSQL compare with a binary collation here, so the row it returns can never be `vm`. The second clause is dead.

It costs nothing to keep, but it tells the next reader that colliding with the current row is possible, and that reader will go looking for the case that makes it happen. Drop it, or say in a comment that it is there in case the outer `name != vm.Name()` guard ever goes away.

## Not asking for a change

The deferred items in your description are fine as deferred. Handler tests landing with #70, the JSON API not taking a rename, the four-times-repeated ownership check waiting for a shared helper, all reasonable, and the reasoning is written down where the next person will find it.

One question rather than a request. On an empty name, `createVirtualModel` redirects to the list page instead of back to `/new`, which drops the description the user typed. The personal side does the same, so it is at least consistent. Deliberate?

Separately, #70 is worth doing soon rather than eventually. It is the same class of hole you just closed, on five middleware handlers, and three of them write.

Bornholm (changes requested):
Reviewed at `abcbd53`. Both points from the previous round are addressed in the code, the cross-org fix is a welcome addition, and CI is green. One of the two fixes does not work, though, and that is what this round is about.

## What is now closed

The dead `existing.ID() != vm.ID()` clause is gone from both update handlers, and the comment you left in its place states the reason correctly: under the outer `name != vm.Name()` guard, `Get...ByName` matches on (org_id, name) or (user_id, name), so a hit can never be the current row. That was the right call between the two options I offered.

The broken type assertion in `updatePersonalModel` is fixed. Asserting `*model.BasePersonalVirtualModel` on a `*wrappedPersonalVirtualModel` would have returned a 500 on every personal rename; the local `mutable` interface makes that unrepresentable.

The four `vm.OrgID() != org.ID()` guards are in place at `virtual_models.go:196`, `:249`, `:345` and `:391`, and `deleteVirtualModel` now resolves the org from the slug before checking. Closing that hole in the same PR was the right thing to do even though it widened the scope.

## The rename race still shows the wrong error

`SaveVirtualModel` (`internal/adapter/gorm/virtual_model_store.go:85`) and `SavePersonalVirtualModel` (`personal_virtual_model_store.go:85`) both test `errors.Is(err, gorm.ErrDuplicatedKey)`. That test can never be true in this codebase.

GORM only produces `ErrDuplicatedKey` when `TranslateError` is set on the config. In `v1.31.2`, `DB.AddError` (gorm.go:407) runs the dialector's translator under `if db.Config.TranslateError` and returns the raw driver error otherwise. `TranslateError` appears nowhere in this repo, and all three `gorm.Open` calls pass a config without it: `internal/setup/gorm_database.go:65`, `cmd/seed/main.go:90`, `test/e2e/harness_test.go:242`.

So the raw `*sqlite3.Error` or `*pgconn.PgError` flows through, the handler's `errors.Is(err, port.ErrAlreadyExists)` branch is never taken, and a user who loses a rename race reads "L'enregistrement a échoué" instead of "ce nom existe déjà". That is the exact symptom the mapping was added to fix.

The repo already has the helper that works. `isUniqueViolation(err, fragments...)` in `internal/adapter/gorm/dialect.go:66` inspects `*sqlite3.Error` and `*pgconn.PgError` (code 23505) directly, with no dependency on `TranslateError`, and four stores already use it: `org_store.go:17`, `role_store.go:21`, `tenant_store.go:17`, `user_store.go:120`.

Swap the condition:

```go
if isUniqueViolation(err, "org_id", "name") {
    return errors.WithStack(port.ErrAlreadyExists)
}
```

and the personal one with `"user_id", "name"`. Same shape as the four existing call sites, one mechanism in the package instead of two.

## Cover the mapping with a store test

This is the reason the bug shipped, so I would rather it not depend on someone remembering. `repository_models_test.go` only exercises the nominal round-trip of `SaveVirtualModel`, and there is no test file at all for the personal model store.

Add an `eachBackend` case: create two virtual models in one org, rename the second onto the first's name through `SaveVirtualModel`, assert `errors.Is(err, port.ErrAlreadyExists)`. Same for `SavePersonalVirtualModel` with two models owned by the same user. Written against the current code, both fail on SQLite and on PostgreSQL. That is the point.

## Fold in if you are touching the file anyway

The `mutable` interface is now declared four times across the repo, identically, at `org/virtual_models.go:291`, `profile/personal_models.go:240` and the two JSON handlers. A named interface in `internal/core/model` or in `webui/common` would stop the next handler from writing a fifth copy. Not blocking.

## Not asking for a change

Deferring the org-ownership helper so it can cover this PR and the middleware handlers of #70 in one pass is sensible; four repeated guards are cheaper to read than a premature abstraction.

Leaving `name` unhandled in the JSON PUT handlers is fine to defer, but it is worth a line in #60 or its own issue: the web UI can now rename and API automation cannot, and that asymmetry is easy to trip over later.

Redirecting to the list page on an empty name loses the typed description. It matches the pre-existing `create_failed` and `exists` behaviour on those two handlers, so I am not asking you to fix it here.

Bornholm (changes requested):
~

Bornholm (changes requested):
## Conclave review, round 3

The two points from round 2 are in. The org guards are there, the type assertion is gone, the handlers translate `ErrAlreadyExists` to `?error=exists`. Ship it.

### Changes requested

**No test covers the collision path in SaveVirtualModel / SavePersonalVirtualModel.**

You asked for this at round 2 and it is still missing. `SaveVirtualModel` (virtual_model_store.go:72-94) and its counterpart for personal models call `isUniqueViolation(err, "org_id"/"user_id", "name")` to turn a driver error into `port.ErrAlreadyExists`. There is no test for this path. The only call to `SaveVirtualModel` in the test suite (`repository_models_test.go:311`) is a nominal round-trip that updates a description. There is no test file for the personal model store.

This already broke once in this PR: at `abcbd53` the code checked `gorm.ErrDuplicatedKey`, which is unreachable. No CI failure, no signal. A later store refactor, or a rename of the index fragment, silently turns "name already exists" into "record update failed".

Add one scenario per store via `eachBackend`: create two models under the same org/user, rename the second to the first's name via `Save*`, assert `errors.Is(err, port.ErrAlreadyExists)`.

### Worth folding in

`CreateVirtualModel` and `CreatePersonalVirtualModel` still check `gorm.ErrDuplicatedKey` (virtual_model_store.go:16, personal_virtual_model_store.go:16). GORM only produces `ErrDuplicatedKey` when `db.Config.TranslateError` is set, and none of the three `gorm.Open` calls in this repo do that. Concurrent creates of the same name return a 500 instead of a 409. Replace with `isUniqueViolation(err, "org_id", "name")` and `isUniqueViolation(err, "user_id", "name")` like every other store in the package.

### Not asking for changes

- **The local `mutable` interface is declared three times** (org/virtual_models.go:291, profile/personal_models.go:240, api/personal_virtual_models.go:151). Non-blocking, noted as a maintainability issue.
- **Cross-org access on middleware handlers** — already tracked in #70, out of scope here.
- **PUT JSON silently ignores the name field** — previously noted, out of scope.

Bornholm (commented):
## Conclave review, round 4

No changes requested. Everything from rounds 2 and 3 is in the tree:

- `SaveVirtualModel` and `SavePersonalVirtualModel` use `isUniqueViolation`.
- `CreateVirtualModel` and `CreatePersonalVirtualModel` also use `isUniqueViolation` — `gorm.ErrDuplicatedKey` is gone from both stores.
- `TestVirtualModelStore_RenameCollisionRejected` and `TestPersonalVirtualModelStore_RenameCollisionRejected` cover the collision path through `eachBackend`, asserting `errors.Is(err, port.ErrAlreadyExists)`.
- The four `vm.OrgID() != org.ID()` guards are in place. `deleteVirtualModel` resolves the org from the slug before checking it.
- The broken type assertion in `updatePersonalModel` is replaced by the local `mutable` interface. The dead `existing.ID() != vm.ID()` clause is gone.

### Out of scope (not blocking merge)

Two findings in files this PR does not touch, marked worth folding in separately:

- **Middleware store still checks `gorm.ErrDuplicatedKey`** (`middleware_store.go:16`) — same pattern this PR fixed in virtual model stores. Concurrent creates of the same name return a 500 instead of a 409.
- **Middleware interface missing `OrgID()`** (`middleware.go:34`) — prerequisite to adding the org guard to middleware handlers. Tracked in #70.

No other in-scope findings.

PR #69, opened 2026-09-18
 feat(models): add price sort on /models page

### LINKED ISSUE #41: Trie par tarif de modèle
### Problem

On peut rapidement griller ses tokens si on ne fait trop attention au prix du token.
Avoir un trie sur https://xolo.cadoles.com/models?show_all=true par tarif pourrait aider à ne pas choisir un modèle trop onéreux 

### Proposed solution

_No response_

### Alternatives considered

_No response_

Padiwa:
### Close #41                                                                                                                                                                                                          
                                                                                                                                                                                                                    
 Add a price-based sort option on the /models page so users can quickly see the cheapest models first and avoid burning tokens on an unexpectedly expensive model.                                                  
                                                                                                                                                                                                                    
 ## What                                                                                                                                                                                                           
                                                                                                                                                                                                                    
 - New URL query parameter sort on /models:                                                                                                                                                                         
     - sort=price → order models by total cost (PromptCostPer1KTokens + CompletionCostPer1KTokens) ascending.                                                                                                       
     - No parameter (or sort=usage) → keep the current default order, by aggregate usage descending.                                                                                                                
 - Virtual models (no direct pricing) are pushed to the end of the list when sorting by price, so they never appear at the top with a misleading "free" position.                                                   
 - New segmented control rendered next to the existing range selector: Usage / Price tabs that toggle the sort while preserving range and show_all query parameters.                                                
                                                                                                                                                                                                                    
 ## Why                                                                                                                                                                                                            
                                                                                                                                                                                                                    
 Users can quickly run through their tokens when they don't know which model costs what. Surfacing a price-sorted view makes it easy to pick an affordable model first. See #41.                                    
                                                                                                                                                                                                                    
 ## How                                                                                                                                                                                                            
                                                                                                                                                                                                                    
 - `internal/http/handler/webui/models.go`: read the new sort query param and extend the sort.Slice comparator. When sort=price, virtual models are sorted last, then pairs are compared by their combined prompt +   
   completion cost; ties fall back to the existing usage comparison.                                                                                                                                                
 - `internal/http/handler/webui/profile/component/models_page.templ`: add a Sort field to ModelsPageVModel and render a new modelsSortControl component (with sortTab child) that builds anchors from CurrentURL via  
   WithoutValues("sort") and WithValues("sort", "price").                                                                                                                                                           
 - `internal/http/handler/webui/common/component/util.go`: add two small helpers (CurrentURLString, MutateURLString) used by the templ component to apply URL mutations across branches.                              
                                                                                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                    
 ### Notes                                                                                                                                                                                                          
                                                                                                                                                                                                                    
 - The /models page is the personal usage page targeted by the issue. The organization-level `/orgs/{slug}/admin/providers/.../models page` is a provider configuration screen and is intentionally untouched.        
 - The price sort remains server-rendered (URL-driven), so it's bookmarkable and works with `?show_all=true`.


Bornholm (commented):
The sort logic is correct and I have nothing to report on behaviour. The comparator is a proper strict weak ordering, the `sort` param is normalised before it reaches the comparator so the active tab can never disagree with the effective order, and dropping `order` before re-adding it is the right way around `httpURL.WithValues` using `query.Add`. Tests cover it well: 17 subtests on the comparator, 4 on the segmented control including the hidden-input round-trip. CI is green.

My one ask is about comment density.

## Cut the commentary back

Across the three hand-written files the PR adds 271 lines, 115 of which are comments. That is 42%, well above the rest of the codebase, and most of it restates the code on the next line or argues with a reader who is not there.

Three examples.

`models.go:32-39` spends eight lines explaining a two-line `if`. It names a hypothetical `?sort=price_invalid`, walks through which tab would render active, then asks a future maintainer not to move the normalisation into the template. The code says `if sortParam != "" && sortParam != "price" { sortParam = "" }`. One line is enough: reject unknown values so the active tab matches the ordering.

The doc comment on `compareModelUsages` runs about 25 lines, and then the body repeats several of its points inline. `// Both are non-virtual (the early return on aVirtual != bVirtual guarantees this)` restates the three lines directly above it. So does the second explanation of the unused-model push at line 175, already covered in the doc comment. Keep the doc comment, drop the echoes.

`sortOrder.String()` is a five-line switch carrying a four-line comment about how the value flows into `ModelsPageVModel.Order`, the toggle icon and the templ href builders. The method returns "asc" or "desc". Nothing else needs saying.

Same pattern in `models_page.templ`: 53 comment lines for 67 lines of code. The `OrderExplicit` field comment explains bookmark pollution over five lines where "true when the request carried an explicit ?order=" would do.

A rough target: keep doc comments on exported and non-obvious functions, drop inline comments that paraphrase the statement they sit on. That should land the diff around 15% comments without losing anything a reader needs.

## Worth keeping

Two blocks earn their space, and I would not want them lost in the cleanup.

The `query.Add` explanation in `usageSegment`. It documents an API trap that is invisible at the call site and that you actually hit, going by the commit history.

The URL contract table above `modelsSortControl`. Four lines that give the whole state machine. That is the comment I would reach for first when coming back to this in six months.

## One comment to fix

The note at `models.go:156` says the early return guarantees both models are non-virtual when entering the `if !aVirtual` block. That reads as if the guard were dead, and it prompted exactly that suggestion during review. It is not dead: after the early return the two flags are equal, and they can both be true. The guard is what keeps two virtual models out of the price comparison. Removing it happens to be harmless today only because `virtualModelAsLLMModel` returns 0 for both cost accessors, so the pair ties and falls through to usage anyway. Keep the guard, and reword the comment to state the precondition it enforces rather than why it looks unnecessary.

## Not asking for a change

Sorting on the per-1K rate rather than `Aggregate.TotalCost` is the right call for #41, since the point is choosing a model before spending on it. `MutateURLString` swallowing a parse error is fine while its only caller derives from `r.URL`; a `slog` line would be cheap if you feel like it, but I am not asking.

Bornholm (changes requested):

